PR #26, opened 2024-12-20
STYLE: Move `math.utils` testing to root test folder

jhlegarreta:
- STYLE: Move `math.utils` testing to root test folder
- ENH: Remove deprecated `math.utils.compute_angle`


jhlegarreta:
Re: d65a001 no tags/releases exist yet, so we are not breaking the API.

jhlegarreta:
Leaving the semidefinite positiveness check function in case we decide to call it from the covariance matrix construction methods.

oesteban:
> Leaving the semidefinite positiveness check function in case we decide to call it from the covariance matrix construction methods.

Scikit-learn checks this internally. I think it can safely go away :)

jhlegarreta:
> Scikit-learn checks this internally. I think it can safely go away :)

Removed.

PR #27, opened 2024-12-20
DOC: Fix package name transition year date

jhlegarreta:
Fix package name transition year date.


jhlegarreta:
If you give me the mecessary rights I can unarchive, apply the change to `eddymotion` and archive it again so that the notice does not become misleading.

PR #28, opened 2024-12-20
MAINT: Add static type checking

jhlegarreta:
Add static type checking: use `mypy` to ensure that variable and function calls are using the appropriate types.

Configure and run `mypy` checks in CI, including the spell checking.

Documentation:
https://mypy.readthedocs.io/en/stable/index.html

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -154,20 +154,20 @@ htmlhelp_basename = "nifreezedoc"
 
 # -- Options for LaTeX output ------------------------------------------------
 
-latex_elements = {
-    # The paper size ('letterpaper' or 'a4paper').
-    #
-    # 'papersize': 'letterpaper',
-    # The font size ('10pt', '11pt' or '12pt').
-    #
-    # 'pointsize': '10pt',
-    # Additional stuff for the LaTeX preamble.
-    #
-    # 'preamble': '',
-    # Latex figure (float) alignment
-    #
-    # 'figure_align': 'htbp',
-}
+# latex_elements = {
+#     # The paper size ('letterpaper' or 'a4paper').
+#     #
+#     # 'papersize': 'letterpaper',
+#     # The font size ('10pt', '11pt' or '12pt').
+#     #
+#     # 'pointsize': '10pt',
+#     # Additional stuff for the LaTeX preamble.
+#     #
+#     # 'preamble': '',
+#     # Latex figure (float) alignment
+#     #
+#     # 'figure_align': 'htbp',
+# }
 
 # Grouping the document tree into LaTeX files. List of tuples
 # (source start file, target name, title,
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -54,7 +54,6 @@ autodoc_mock_imports = [
     "nipype",
     "nitime",
     "nitransforms",
-    "numpy",
     "pandas",
     "scipy",
     "seaborn",
--- a/src/nifreeze/cli/parser.py
+++ b/src/nifreeze/cli/parser.py
@@ -29,13 +29,13 @@ from typing import Optional
 import yaml
 
 
-def _parse_yaml_config(file_path: Path) -> dict:
+def _parse_yaml_config(file_path: str) -> dict:
     """
     Parse YAML configuration file.
 
     Parameters
     ----------
-    file_path : Path
+    file_path : str
         Path to the YAML configuration file.
 
     Returns
--- a/src/nifreeze/model/gpr.py
+++ b/src/nifreeze/model/gpr.py
@@ -212,7 +212,7 @@ class DiffusionGPR(GaussianProcessRegressor):
     ) -> tuple[float, float]:
         options = {}
         if self.optimizer == "fmin_l_bfgs_b":
-            from sklearn.utils.optimize import _check_optimize_result
+            from sklearn.utils.optimize import _check_optimize_result  # type: ignore
 
             for name in LBFGS_CONFIGURABLE_OPTIONS:
                 if (value := getattr(self, name, None)) is not None:
--- a/src/nifreeze/model/gpr.py
+++ b/src/nifreeze/model/gpr.py
@@ -25,7 +25,7 @@
 from __future__ import annotations
 
 from numbers import Integral, Real
-from typing import Callable, ClassVar, Mapping, Optional, Sequence, Union
+from typing import Callable, ClassVar, Literal, Mapping, Optional, Sequence, Union
 
 import numpy as np
 from scipy import optimize
@@ -171,7 +171,7 @@ class DiffusionGPR(GaussianProcessRegressor):
         kernel: Kernel | None = None,
         *,
         alpha: float = 0.5,
-        optimizer: str | Callable | None = "fmin_l_bfgs_b",
+        optimizer: Literal["fmin_l_bfgs_b"] | Callable | None = "fmin_l_bfgs_b",
         n_restarts_optimizer: int = 0,
         copy_X_train: bool = True,
         normalize_y: bool = True,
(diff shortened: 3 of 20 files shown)


jhlegarreta:
Unsure for now if the following need to be added to the `tool.mypy.overrides` list:
- `dipy`
- `scikit-image`
- `scikit_learn`
- `scipy`

@effigies I see that in this repository the `style` checking is being done through an explicit call to `ruff` in a dedicated GHA workflow, as opposed to the way it is done in `nireports` which I watched as the inspiration for this PR:
https://github.com/nipreps/nifreeze/blob/7a481188d85f3e34b47bb1327383f229643ddb9a/.github/workflows/contrib.yml#L22-L25
vs
https://github.com/nipreps/nireports/blob/main/tox.ini#L72
https://github.com/nipreps/nireports/blob/main/.github/workflows/build_test_deploy.yml#L159
https://github.com/nipreps/nireports/actions/runs/12416179089/job/34664200380

The latter seems to be more elegant and self-contained. Is there a canonical way to do this through the `NiPreps`?

jhlegarreta:
Looks like the `ignore_missing_imports` flag is not being considered:
https://github.com/nipreps/nifreeze/pull/28/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R141
e.g.
https://github.com/nipreps/nifreeze/actions/runs/12437972140/job/34728973936?pr=28#step:8:48

I would have thought that `mypy` would consider it somehow when executing:
https://github.com/nipreps/nifreeze/pull/28/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R56

Any clue @effigies ?

jhlegarreta:
Spent a few hours on this; solved all those warnings that I could: Down to 10 locally using the `--ignore-missing-import` flag. A few notes:
- The error
  ```
  src/nifreeze/model/gpr.py:187: error:
  Unexpected keyword argument "n_targets" for "__init__" of "GaussianProcessRegressor"  [call-arg]
  ```
  seems legitimate: we are requiring `scikit-learn>=0.18` (a pretty old one) and `n_targets` did not appear until `scikit_learn==1.3.0`. However, testing is using `scikit-learn 1.6.0`:
  https://github.com/nipreps/nifreeze/actions/runs/12437972140/job/34728973936#step:8:29
  which does have it https://scikit-learn.org/1.6/modules/generated/sklearn.gaussian_process.GaussianProcessRegressor.html
  So not sure what's going on here.
- I'd dare to say that the `NiBabel` `load` function seems to have some misleading type hints:
  https://github.com/nipy/nibabel/blob/59733ff5a9eedb73e63261a1d706bfe1efbf0f50/nibabel/loadsave.py#L82
  which results in `mypy` complaining, e.g.
  https://github.com/nipreps/nifreeze/actions/runs/12437972140/job/34728973936#step:8:63:
  ```
  scripts/dwi_gp_estimation_signal_plot.py:113: error:
  "FileBasedImage" has no attribute "get_fdata"  [attr-defined]
  ```
  https://github.com/nipreps/nifreeze/actions/runs/12437972140/job/34728973936#step:8:142:
  ```
  scripts/optimize_registration.py:173: error:
  Argument 1 to "displacements_within_mask" has incompatible type "FileBasedImage";
  expected "SpatialImage"  [arg-type]
  ```
  Have tried to pin this down in `NiBabel` without success. Am I missing something @effigies? Can push a patch to `NiBabel` if you give me some context/doable for a non-expert.
- Preferred to do separate PRs for these as they involve more than just type hints: #29, #30

jhlegarreta:
Also, looks like scripts are not being considered: https://github.com/nipreps/nifreeze/pull/41. Any guess why?

README.rst
oesteban, line 1: Let's correct this on a separate PR, it's mixing up a typo with type checking.

docs/conf.py
oesteban, line 1: Perhaps comment out the block instead of deleting?

src/nifreeze/cli/parser.py
```diff
@@ -29,13 +29,13 @@
 import yaml
 
 
-def _parse_yaml_config(file_path: Path) -> dict:
+def _parse_yaml_config(file_path: str) -> dict:
```
oesteban, line 32:
```suggestion
def _parse_yaml_config(file_path: str | Path) -> dict:
```

src/nifreeze/model/gpr.py
```diff
@@ -332,7 +334,7 @@ def __call__(
 
         return self.beta_l * C_theta, K_gradient
 
-    def diag(self, X: np.ndarray) -> np.ndarray:
```
oesteban, line 335: ?

src/nifreeze/testing/simulations.py
oesteban, line 1: For a separate PR with typos in documentation?

oesteban:
> * seems legitimate: we are requiring `scikit-learn>=0.18` (a pretty old one) and `n_targets` did not appear until `scikit_learn==1.3.0`. However, testing is using `scikit-learn 1.6.0`:

We should then upgrade the pinning.

README.rst
jhlegarreta: PR #49.

src/nifreeze/testing/simulations.py
jhlegarreta: PR #49.

src/nifreeze/cli/parser.py
jhlegarreta: I do think the `Path` case will be ever hit: https://github.com/nipreps/nifreeze/commit/0a313000bf2ac35fce040055f86cc38a153854d7

src/nifreeze/model/gpr.py
jhlegarreta: https://github.com/nipreps/nifreeze/pull/28/commits/8341c3087ffb1709f202ad274452dcd5c81cb9e4

effigies (commented):
mypy is now passing on this project. I think the changes I made were in the spirit of what is intended here.

effigies (approved):
I've reviewed the whole thing, including my own contributions, and I'm good with this. It could probably do with another pair of eyes.

src/nifreeze/model/gpr.py
effigies: Fixed in bd38772.

jhlegarreta: Thanks @effigies !

jhlegarreta:
@effigies you rock, thanks for doing all this :100:.

The changes to the iterators were being addressed in https://github.com/nipreps/nifreeze/pull/29, and they were in a deadlock state due to this:
https://github.com/nipreps/nifreeze/pull/29/files#r1915664709

So I fear that your change in
https://github.com/nipreps/nifreeze/pull/28/commits/d418e7dcd411f41bb329d62c7d8e1ce6b3afc79f#diff-196825dee04c087a145cee5f8533c4668eefe92e28f43d33200bbea651643edeR56

brings us back to the question. I would prefer that we solved this in PR #29, and then rebase this one. I can take care of that once we agree on what the best solution is for the question here.

Besides that, I have taken the freedom to remove the add/revert of building the documentation with Python 3.12 to keep the history cleeaner and I have removed this
https://github.com/nipreps/nifreeze/pull/28/commits/d2ef82f682aba4a4419e9410e918efd239ba40dd#diff-d26d09b32ad847acc6e6ab4846378088e1c8c072df896e53b76f682fddedfd2bR196

as it was fixed by https://github.com/nipreps/nifreeze/pull/58.

The `ruff` failure is being addressed in PR #64.

oesteban:
This is such a big PR I fear we will not be able to ever finish it if we stop at every rock on the way. Let's merge this one in and then resolve side effects and nuances (#29, #30, #64) later.

PR #29, opened 2024-12-22
DOC: Revision of docstrings in the iterators' module

jhlegarreta:
- ENH: Add None to b-vals iterators `size` parameter annotations
- ENH: Make `linear_iterator` return an iterator
- ENH: Remove unused parameter form `bvalue_iterator` prototype

### DIFF
--- a/src/nifreeze/utils/iterators.py
+++ b/src/nifreeze/utils/iterators.py
@@ -33,9 +33,11 @@ def linear_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
 
     Parameters
     ----------
-    size : :obj:`int` or ``None`` (optional)
+    size : :obj:`int` or ``None``, optional
         Number of volumes in the dataset.
-    bvalues : :obj:`list`
+        If ``None``, ``size`` will be inferred from the ``bvalues``
+        keyword argument.
+    bvalues : :obj:`list`, optional (keyword argument)
         List of b-values corresponding to all orientations of a DWI dataset.
         If ``size`` is provided, this argument will be ignored.
         Otherwise, ``size`` will be inferred from the length of ``bvalues``.
@@ -63,20 +65,22 @@ def random_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
     """
     Traverse the dataset volumes randomly.
 
-    If the `seed` key is present in the keyword arguments, initializes the seed
-    of Python's `random` pseudo-random number generator library with the given
-    value. Specifically, if `False`, `None` is used as the seed; it `True`, a
+    If the ``seed`` key is present in the keyword arguments, initializes the seed
+    of Python's ``random`` pseudo-random number generator library with the given
+    value. Specifically, if ``False``, ``None`` is used as the seed; if ``True``, a
     default seed value is used.
 
     Parameters
     ----------
-    size : :obj:`int` or ``None`` (optional)
+    size : :obj:`int` or ``None``, optional
         Number of volumes in the dataset.
-    bvalues : :obj:`list` (optional, keyword argument)
+        If ``None``, ``size`` will be inferred from the ``bvalues``
+        keyword argument.
+    bvalues : :obj:`list`, optional (keyword argument)
         List of b-values corresponding to all orientations of a DWI dataset.
         If ``size`` is provided, this argument will be ignored.
         Otherwise, ``size`` will be inferred from the length of ``bvalues``.
-    seed : :obj:`int` or :obj:`bool` or :obj:`bool` or ``None`` (optional, keyword argument)
+    seed : :obj:`int`, :obj:`bool`, :obj:`str`, or ``None``, optional (keyword argument)
         If :obj:`int` or :obj:`str` or ``None``, initializes the seed of Python's random generator
         with the given value.
         If ``False``, the random generator is passed ``None``.
@@ -124,6 +128,8 @@ def bvalue_iterator(*_, **kwargs) -> Iterator[int]:
     ----------
     bvalues : :obj:`list`
         List of b-values corresponding to all orientations of the dataset.
+        Please note that ``bvalues`` is a keyword argument and MUST be provided
+        to generate the volume sequence.
 
     Yields
     ------
@@ -149,9 +155,11 @@ def centralsym_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
 
     Parameters
     ----------
-    size : :obj:`int` or ``None`` (optional)
+    size : :obj:`int` or ``None``, optional
         Number of volumes in the dataset.
-    bvalues : :obj:`list` (optional, keyword argument)
+        If ``None``, ``size`` will be inferred from the ``bvalues``
+        keyword argument.
+    bvalues : :obj:`list`, optional (keyword argument)
         List of b-values corresponding to all orientations of the dataset.
         If ``size`` is provided, this argument will be ignored.
         Otherwise, ``size`` will be inferred from the length of ``bvalues``.
--- a/src/nifreeze/utils/iterators.py
+++ b/src/nifreeze/utils/iterators.py
@@ -107,14 +107,9 @@ def random_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
     return (x for x in index_order)
 
 
-def bvalue_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
+def bvalue_iterator(*_, **kwargs) -> Iterator[int]:
     """
     Traverse the volumes in a DWI dataset by growing b-value.
-
-    Parameters
-    ----------
-    bvalues : :obj:`list`
-        List of b-values corresponding to all orientations of the dataset.
 
     Returns
     -------
--- a/src/nifreeze/utils/iterators.py
+++ b/src/nifreeze/utils/iterators.py
@@ -33,14 +33,16 @@ def linear_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
 
     Parameters
     ----------
-    size : :obj:`int` or None, optional
-        Number of volumes in the dataset (for instance, the number of
-        orientations in a DWI). If `None`, the size is inferred from the `bvals`
-        argument in `kwargs`, if present.
+    size : :obj:`int` or ``None`` (optional)
+        Number of volumes in the dataset.
+    bvalues : :obj:`list`
+        List of b-values corresponding to all orientations of a DWI dataset.
+        If ``size`` is provided, this argument will be ignored.
+        Otherwise, ``size`` will be inferred from the length of ``bvalues``.
 
-    Returns
-    -------
-    :obj:`~typing.Iterator`
+    Yields
+    ------
+    :obj:`int`
         The sorted index order.
 
     Examples
@@ -54,7 +56,7 @@ def linear_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
     if size is None:
         raise TypeError("Cannot build iterator without size")
 
-    return iter(range(size))
+    return (s for s in range(size))
 
 
 def random_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
@@ -68,14 +70,21 @@ def random_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
 
     Parameters
     ----------
-    size : :obj:`int` or None, optional
-        Number of volumes in the dataset (for instance, the number of
-        orientations in a DWI). If `None`, the size is inferred from the `bvals`
-        argument in `kwargs`, if present.
+    size : :obj:`int` or ``None`` (optional)
+        Number of volumes in the dataset.
+    bvalues : :obj:`list` (optional, keyword argument)
+        List of b-values corresponding to all orientations of a DWI dataset.
+        If ``size`` is provided, this argument will be ignored.
+        Otherwise, ``size`` will be inferred from the length of ``bvalues``.
+    seed : :obj:`int` or :obj:`bool` or :obj:`bool` or ``None`` (optional, keyword argument)
+        If :obj:`int` or :obj:`str` or ``None``, initializes the seed of Python's random generator
+        with the given value.
+        If ``False``, the random generator is passed ``None``.
+        If ``True``, a default seed value is set.
 
-    Returns
-    -------
-    :obj:`~typing.Iterator`
+    Yields
+    ------
+    :obj:`int`
         The sorted index order.
 
     Examples
@@ -111,9 +120,14 @@ def bvalue_iterator(*_, **kwargs) -> Iterator[int]:
     """
     Traverse the volumes in a DWI dataset by growing b-value.
 
-    Returns
-    -------
-    :obj:`~typing.Iterator`
+    Parameters
+    ----------
+    bvalues : :obj:`list`
+        List of b-values corresponding to all orientations of the dataset.
+
+    Yields
+    ------
+    :obj:`int`
         The sorted index order.
 
     Examples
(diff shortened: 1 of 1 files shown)


jhlegarreta:
Note this change:
https://github.com/nipreps/nifreeze/pull/29/files#diff-caf7665e7d0c7d0454e8984b460aa419f31154c8746083014c1524d306ab3644R110

oesteban (approved):

src/nifreeze/utils.py
```diff
@@ -53,23 +54,24 @@ def linear_iterator(size: int = None, **kwargs) -> Iterator[int]:
     if size is None:
         raise TypeError("Cannot build iterator without size")
 
-    return range(size)
+    return iter(range(size))
```
oesteban, line 57: isn't `range()` a generator itself?
```suggestion
    return range(size)
```

oesteban (approved):
I don't see why the `iter()` decoration to range, other than that looks good!

jhlegarreta:
> There seems to be some conflicts after the style update I made.

Force pushed; conflicts gone.

src/nifreeze/utils.py
oesteban: Sorry, I had always assumed that `range` is a generator. But it is not, its of type `range`.

I can't suggest here, adding a new comment.

src/nifreeze/utils/iterators.py
```diff
@@ -53,23 +54,24 @@ def linear_iterator(size: int = None, **kwargs) -> Iterator[int]:
     if size is None:
         raise TypeError("Cannot build iterator without size")
 
-    return range(size)
+    return iter(range(size))
```
oesteban, line 57: Like for the other iterators, let's return a generator:

```suggestion
    return (s for s in range(size))
```

oesteban (approved):

src/nifreeze/utils/iterators.py
```diff
@@ -53,23 +54,24 @@ def linear_iterator(size: int = None, **kwargs) -> Iterator[int]:
...
-        (for instance, the number of orientations in a DWI)
-    seed : :obj:`int` or :obj:`bool` or :obj:`bool` or ``None``
-        If :obj:`int` or :obj:`str` or ``None``, initializes the seed of Python's random generator
-        with the given value.
-        If ``False``, the random generator is passed ``None``.
-        If ``True``, a default seed value is set.
+    size : :obj:`int` or None, optional
+        Number of volumes in the dataset (for instance, the number of
+        orientations in a DWI). If `None`, the size is inferred from the `bvals`
+        argument in `kwargs`, if present.
```
oesteban, line 74: Although it is not explicitly stated in the signature, I think we can use the opportunity that sphinx allows us to document seed here if we want. Removing the specification of seed here makes this documentation less accessible and useful.
```suggestion
    size : :obj:`int` or None, optional
        Number of volumes in the dataset (for instance, the number of
        orientations in a DWI). If `None`, the size is inferred from the `bvals`
        argument in `kwargs`, if present.
    seed : :obj:`int` or :obj:`bool` or :obj:`bool` or ``None``
        If :obj:`int` or :obj:`str` or ``None``, initializes the seed of Python's random generator
        with the given value.
        If ``False``, the random generator is passed ``None``.
        If ``True``, a default seed value is set.
```

src/nifreeze/utils/iterators.py
```diff
@@ -105,15 +107,10 @@ def random_iterator(size: int = None, **kwargs) -> Iterator[int]:
...
-def bvalue_iterator(size: int = None, **kwargs) -> Iterator[int]:
+def bvalue_iterator(*_, **kwargs) -> Iterator[int]:
     """
     Traverse the volumes in a DWI dataset by growing b-value.
 
-    Parameters
-    ----------
-    bvalues : :obj:`list`
-        List of b-values corresponding to all orientations of the dataset.
-
```
oesteban, line 116: Please roll back, as for seed above, it is worthwhile documenting keyword args here.

src/nifreeze/utils/iterators.py
jhlegarreta: We should maybe spare some more thought to this: the functions being named `*_iterator` is misleading now, and the return type docstrings are misleading as well, as they all say `:obj:~typing.Iterator`. To avoid too many back and forths, and as I have not fully understood https://github.com/nipreps/nifreeze/pull/29#discussion_r1916292026, better to discuss this live over a call.

src/nifreeze/utils/iterators.py
oesteban: As per numpydoc guidelines, [this seems to be a use case for *Other parameters*](https://numpydoc.readthedocs.io/en/latest/format.html#other-parameters). Updated the suggestion.

oesteban: That said, I don't think it clutters the parameters section in this case, so I would just leave it with the parameters section.

oesteban:
In 6398249, I have added the following changes:

* Revert the removal of optional keyword arguments from docstrings. The rationale is that, if they are not documented there, the type hints in the signature are insufficient to introspect what keyword arguments may be provided and their behavior. As [discussed above](https://github.com/nipreps/nifreeze/pull/29#discussion_r1926521505), you can use [`Other parameters` for this as per numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#other-parameters). However, in this particular case I find it overly verbose. Instead, I propose we explicitly label these arguments as *optional, keyword argument*.
* Rewrite the *Returns* section with *Yields*, following [numpydocs' documentation](https://numpydoc.readthedocs.io/en/latest/format.html#yields), as all these are meant to yield values rather than return values.

In addition to those changes, I also rolled back the `iter()` decoration of the linear iterator, and replace it with an actual generator, for consistency with the other iterators.

You can always convert these generators to iterators or lists at any moment.

@jhlegarreta @effigies let me know if the above sounds reasonable.

src/nifreeze/utils/iterators.py
```diff
@@ -33,14 +33,17 @@ def linear_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
 
     Parameters
     ----------
-    size : :obj:`int`
-        Number of volumes in the dataset
-        (for instance, the number of orientations in a DWI)
+    size : :obj:`int` or ``None`` (optional)
```
jhlegarreta, line 36: size : :obj:`int` or ``None`` (optional)
```suggestion
    size : :obj:`int` or ``None``, optional
```

src/nifreeze/utils/iterators.py
```diff
@@ -53,28 +56,36 @@ def linear_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
...
 
 
 def random_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
     """
     Traverse the dataset volumes randomly.
 
+    If the `seed` key is present in the keyword arguments, initializes the seed
+    of Python's `random` pseudo-random number generator library with the given
+    value. Specifically, if `False`, `None` is used as the seed; it `True`, a
+    default seed value is used.
```
jhlegarreta, line 69:
```suggestion
    If the ``seed`` key is present in the keyword arguments, initializes the seed
    of Python's ``random`` pseudo-random number generator library with the given
    value. Specifically, if ``False``, ``None`` is used as the seed; if ``True``, a
    default seed value is used.
```

src/nifreeze/utils/iterators.py
```diff
@@ -53,28 +56,36 @@ def linear_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
...
+
     Parameters
     ----------
-    size : :obj:`int`
-        Number of volumes in the dataset
-        (for instance, the number of orientations in a DWI)
-    seed : :obj:`int` or :obj:`bool` or :obj:`bool` or ``None``
+    size : :obj:`int` or ``None`` (optional)
```
jhlegarreta, line 73:
```suggestion
    size : :obj:`int` or ``None``, optional
```

src/nifreeze/utils/iterators.py
```diff
@@ -33,14 +33,17 @@ def linear_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
 
     Parameters
     ----------
-    size : :obj:`int`
-        Number of volumes in the dataset
-        (for instance, the number of orientations in a DWI)
+    size : :obj:`int` or ``None`` (optional)
+        Number of volumes in the dataset.
```
jhlegarreta, line 37: Explain the behavior when value is ``None``?

src/nifreeze/utils/iterators.py
```diff
@@ -33,14 +33,17 @@ def linear_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
...
+    bvalues : :obj:`list`
```
jhlegarreta, line 38: I assume this is contained in the kwargs: not sure what the standard way to document this is. In any case, it should be marked as optional or something else, as it comes after an optional argument.

src/nifreeze/utils/iterators.py
```diff
@@ -53,28 +56,36 @@ def linear_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
...
+    default seed value is used.
+
     Parameters
     ----------
-    size : :obj:`int`
-        Number of volumes in the dataset
-        (for instance, the number of orientations in a DWI)
-    seed : :obj:`int` or :obj:`bool` or :obj:`bool` or ``None``
+    size : :obj:`int` or ``None`` (optional)
+        Number of volumes in the dataset.
```
jhlegarreta, line 74: Explain the behavior when using None?

src/nifreeze/utils/iterators.py
```diff
@@ -53,28 +56,36 @@ def linear_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
...
+    bvalues : :obj:`list` (optional, keyword argument)
```
jhlegarreta, line 75: Same comment as above, not sure what Python's standard way to document these is.

src/nifreeze/utils/iterators.py
```diff
@@ -53,28 +56,36 @@ def linear_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
...
+        List of b-values corresponding to all orientations of a DWI dataset.
+        If ``size`` is provided, this argument will be ignored.
+        Otherwise, ``size`` will be inferred from the length of ``bvalues``.
+    seed : :obj:`int` or :obj:`bool` or :obj:`bool` or ``None`` (optional, keyword argument)
```
jhlegarreta, line 79: Same as above.

Additional minor stylistic comment: does Python use:
```
obj:`int` or :obj:`bool` or ``None`` 
```

vs
```
obj:`int`, :obj:`bool` or ``None`` 
```

or something else?

Note that there is a spare :obj:\`bool\`; according to the docstring it should be maybe :obj:\`str\`?

src/nifreeze/utils/iterators.py
```diff
@@ -138,14 +149,17 @@ def centralsym_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
...
     ----------
-    size : :obj:`int`
-        Number of volumes in the dataset
-        (for instance, the number of orientations in a DWI)
-
-    Returns
-    -------
-    :obj:`~typing.Iterator`
-        The sorted index order.
+    size : :obj:`int` or ``None`` (optional)
```
jhlegarreta, line 152:
```suggestion
    size : :obj:`int` or ``None``, optional
```

src/nifreeze/utils/iterators.py
```diff
@@ -138,14 +149,17 @@ def centralsym_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
...
+        Number of volumes in the dataset.
```
jhlegarreta, line 153: Same comment as above.

src/nifreeze/utils/iterators.py
```diff
@@ -114,10 +125,10 @@ def bvalue_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
     bvalues : :obj:`list`
         List of b-values corresponding to all orientations of the dataset.
```
jhlegarreta, line 126: Same comment as above: noting this is an optional/kwarg, and documenting the behavior when it is not provided. There is other examples of the latter, so if the behavior is the same, a simple C&P would work.

src/nifreeze/utils/iterators.py
```diff
@@ -138,14 +149,17 @@ def centralsym_iterator(size: int | None = None, **kwargs) -> Iterator[int]:
...
-        Number of volumes in the dataset
-        (for instance, the number of orientations in a DWI)
-
-    Returns
-    -------
-    :obj:`~typing.Iterator`
-        The sorted index order.
+    size : :obj:`int` or ``None`` (optional)
+        Number of volumes in the dataset.
+    bvalues : :obj:`list` (optional, keyword argument)
```
jhlegarreta, line 154: Same comment as above about the style to document a kwarg.

jhlegarreta (commented):
> In 6398249, I have added the following changes:

> - Revert the removal of optional keyword arguments from docstrings. The rationale is  that, if they are not documented there, the type hints in the signature are insufficient  to introspect what keyword arguments may be provided and their behavior. As  discussed above, you can use Other parameters for this as per numpydoc. However, in this particular case I find it overly verbose. Instead, I propose we explicitly label these arguments as optional, keyword argument.

Agreed that documenting these is useful in this case. Left a few style comment to make documenting these nicer/more Pythonin/Sphinxier if there is a way.

> - Rewrite the Returns section with Yields, following numpydocs' documentation, as all  these are meant to yield values rather than return values.

:+1: `Yield` looks the right thing.

>  In addition to those changes, I also rolled back the iter() decoration of the linear  iterator, and replace it with an actual generator, for consistency with the other > iterators.

OK.

> @jhlegarreta @effigies let me know if the above sounds reasonable.

Changes look OK. The only inconsistency is the `_iterator` naming vs. them being generators.

src/nifreeze/utils/iterators.py
oesteban:
```suggestion
        Number of volumes in the dataset.
        If ``None``, ``size`` will be inferred from the ``bvalues``
        keyword argument.
```

src/nifreeze/utils/iterators.py
oesteban:
```suggestion
    bvalues : :obj:`list`, optional (keyword argument)
```

src/nifreeze/utils/iterators.py
oesteban:
```suggestion
        Number of volumes in the dataset.
        If ``None``, ``size`` will be inferred from the ``bvalues``
        keyword argument.
```

src/nifreeze/utils/iterators.py
oesteban:
```suggestion
    seed : :obj:`int`, :obj:`bool`, :obj:`str`, or ``None``, optional (keyword argument)
```

src/nifreeze/utils/iterators.py
oesteban:
```suggestion
        List of b-values corresponding to all orientations of the dataset.
        Please note that ``bvalues`` is a keyword argument and MUST be provided
        to generate the volume sequence.
```

src/nifreeze/utils/iterators.py
oesteban:
```suggestion
    bvalues : :obj:`list`, optional (keyword argument)
```

src/nifreeze/utils/iterators.py
oesteban:
```suggestion
        Number of volumes in the dataset.
        If ``None``, ``size`` will be inferred from the ``bvalues``
        keyword argument.
```

oesteban:
> The only inconsistency is the `_iterator` naming vs. them being generators.

Let's then punt this inconsistency for the future, as the module is also called "iterators". They are indeed iterators (that's why the recommendation for the type hint of the return is Iterator), however they are not implemented exactly like proper Python iterators.

If the naming is too irritating for you or anyone, happy to see a future PR addressing this.

PR #30, opened 2024-12-22
ENH: Fix GP estimation error analysis script

jhlegarreta:
- ENH: Fix return value type annotation in GP error analysis script func
- ENH: Use a boolean to tell whether to serialize the pandas df index
- BUG: Provide missing position arg to local cross validation function
- ENH: Add type annotation for variable in GP estimation error analysis


jhlegarreta:
Marking as draft:
- This issue is still not solved:
  https://github.com/nipreps/nifreeze/actions/runs/12437972140/job/34728973936#step:8:110
  ```
  scripts/dwi_gp_estimation_error_analysis.py:77: error: Argument "cv" to "cross_val_score" has incompatible type 
  "RepeatedKFold"; expected "int | BaseCrossValidator | Iterable[Any] | BaseShuffleSplit | None"  [arg-type]
  ```

  and related to that, this is probably not what was planned when using `RepeatedKFold`:
  https://github.com/nipreps/nifreeze/pull/30/files#diff-e16a8632d36e294b24752d0ef9e6ea37a63e390a26045e188d8d5d4d0a9dc8ceR206

jhlegarreta:
Most of these changes were addressed in PR #28. The only missing part is that the complaint in https://github.com/nipreps/nifreeze/pull/30#issuecomment-2558297361 seems legitimate, so ignoring the warning https://github.com/nipreps/nifreeze/pull/28/files#diff-e16a8632d36e294b24752d0ef9e6ea37a63e390a26045e188d8d5d4d0a9dc8ceR83 may not be the best solution. We are likely to revamp this script/the scripts as we make progress, so we will revisit this at due time. Keeping it open as a reminder.

PR #34, opened 2024-12-22
ENH: GP hyperparameter selection by cross-validation

jhlegarreta:


jhlegarreta:
Transfer from https://github.com/nipreps/eddymotion/pull/250.

PR #37, opened 2024-12-22
ENH: Warn users if masks are not provided for registration and fitting

### LINKED ISSUE #10: Show a big-fat warning when a brain mask is not provided
With the development of nipreps/eddymotion#91, it has become clear that we don't want to be fitting models in background pixels, nor have the background be considered by the cost function of ANTs.

Therefore, eddymotion should be very reluctant to go ahead without a brain mask.

Incidentally, I believe can easily include SynthStrip into dMRIPrep and then dilate that mask so that distortions do not cause cropping of the brain.

WDYT?

jhlegarreta:
Warn users if masks are not provided for registration and model fitting.

Closes #10.

### DIFF
--- a/src/nifreeze/registration/ants.py
+++ b/src/nifreeze/registration/ants.py
@@ -382,12 +382,16 @@ def generate_command(
         settings["fixed_image_masks"] = [
             str(p) for p in _massage_mask_path(fixedmask_path, nlevels)
         ]
+    else:
+        warn("No fixed image mask provided; consider using a mask to avoid optimization issues.")
 
     # Set moving masks if provided
     if movingmask_path is not None:
         settings["moving_image_masks"] = [
             str(p) for p in _massage_mask_path(movingmask_path, nlevels)
         ]
+    else:
+        warn("No moving image mask provided; consider using a mask to avoid optimization issues.")
 
     # Set initalizing affine if provided
     if init_affine is not None:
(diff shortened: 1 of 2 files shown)


jhlegarreta:
Will add testing whenever we have some testing data.

jhlegarreta:
This is ready to be merged.

src/nifreeze/registration/ants.py
```diff
@@ -382,12 +382,16 @@ def generate_command(
         settings["fixed_image_masks"] = [
             str(p) for p in _massage_mask_path(fixedmask_path, nlevels)
         ]
+    else:
+        warn("No fixed image mask provided; consider using a mask to avoid optimization issues.")
```
oesteban, line 386:
```suggestion
```

src/nifreeze/registration/ants.py
```diff
@@ -382,12 +382,16 @@ def generate_command(
...
 
     # Set moving masks if provided
     if movingmask_path is not None:
         settings["moving_image_masks"] = [
             str(p) for p in _massage_mask_path(movingmask_path, nlevels)
         ]
+    else:
+        warn("No moving image mask provided; consider using a mask to avoid optimization issues.")
```
oesteban, line 394:
```suggestion
```

oesteban (commented):
I think the masks at registration time should not be warned. For instance, if running an initialization level you may want to drop the mask. Since the model is what we care about, I'd leave it only there.

oesteban (approved):

PR #39, opened 2024-12-22
ENH: Use keyword arguments when calling  DIPY keyword arg functions

jhlegarreta:
Pass parameters as keyword arguments when calling DIPY functions and methods that have keyword arguments: DIPY has transitioned to keyword-only arguments for these starting version 1.10.0 (released Dec 12, 2024):
https://docs.dipy.org/1.10.0/api_changes.html#dipy-1-10-0-changes

Fixes:
```
src/nifreeze/testing/simulations.py:181:
 UserWarning: Pass ['bvecs'] as keyword args.
 From version 2.0.0 passing these as positional arguments will result in an error.
      return gradient_table(bvals, bvecs)
```

raised for example in:
https://github.com/nipreps/nifreeze/actions/runs/12457578962/job/34772277392?pr=35#step:12:1015

### DIFF
--- a/scripts/dwi_gp_estimation_signal_plot.py
+++ b/scripts/dwi_gp_estimation_signal_plot.py
@@ -114,7 +114,7 @@ def main() -> None:
     y_pred = nib.load(args.dwi_pred_data_fname).get_fdata()
 
     bvals, bvecs = read_bvals_bvecs(str(args.bval_data_fname), str(args.bvec_data_fname))
-    gtab = gradient_table(bvals, bvecs)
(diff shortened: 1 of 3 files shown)


jhlegarreta:
This is ready to be merged. No min required version bump is necessary:
https://github.com/nipreps/nifreeze/blob/6fd01a3f06a2e0a49e84fbdde265ce747f1fede6/pyproject.toml#L23
https://github.com/dipy/dipy/blob/1.3.0/dipy/core/gradients.py#L475

PR #40, opened 2024-12-22
ENH: Consider warnings as errors when running `pytest`

jhlegarreta:
Consider warnings as errors when running `pytest`:
- Add the `warnings-as-errors` flag to `pytest` to consider warnings as errors.
- Use an environment variable `NIFREEZE_WERRORS` in GHA workflow files to mark the corresponding workflows as treating warnings as errors.
- Process the `pytest` session end to mark it as failed if warnings were present.

Implementation drawn from DIPY.

Filter two warnings that are being raised by the current tests:
- Filter warnings related to GPR parameter optimal values being close to the boundary values raised by `scikit-learn`.
- Filter warnings related to the b0 threshold value being updated in gradient table slicing raised by DIPY.


jhlegarreta:
We are missing these:
https://github.com/nipreps/nifreeze/actions/runs/12434980215/job/34719815043#step:11:1010

jhlegarreta:
Warnings are being filtered, but the test is not being marked as failed:

nifreeze_ci_filtered_not_failed

Marking this as draft.

oesteban (approved):

PR #42, opened 2024-12-23
ENH: Test scripts can be called with the help flag

jhlegarreta:
Test scripts can be called with the help flag.


jhlegarreta:
Tests are failing because the scripts are not being listed in the appropriate section in the `pyproject.toml`. To be revisited once we agree which should go or whether all should be listed for now and be changed as we move forward.

jhlegarreta:
So these scripts were written to serve as a support for generating results, plots, etc. but not to be available to the users _per se_. So their nature is more like a notebook, without the inconveniences of the notebooks in development. Maybe we should convert them into notebooks once we are done with the generation of results, plots, etc. and put them into the `docs/notebooks` folder. So I'd keep the PR as a draft for now.

WDYT @oesteban?

PR #43, opened 2024-12-23
STYLE: Use the same alias consistently to import `nibabel`

jhlegarreta:
Use the same alias consistently to import `nibabel`: prefer using `nb` over `nib` as it is more popular across `nipreps` (used in 273 files vs. 13 as of Jan 8, 2025).

### DIFF
--- a/docs/notebooks/dwi_gp_estimation.ipynb
+++ b/docs/notebooks/dwi_gp_estimation.ipynb
@@ -24,7 +24,7 @@
    "outputs": [],
    "source": [
     "import dipy.data as dpd\n",
-    "import nibabel as nib\n",
+    "import nibabel as nb\n",
     "import numpy as np\n",
     "from dipy.core.gradients import get_bval_indices\n",
     "from dipy.io import read_bvals_bvecs\n",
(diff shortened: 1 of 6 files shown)


effigies:
We predominantly use `import nibabel as nb` throughout nipreps:

* [`/import nibabel as nb/`](https://github.com/search?q=org%3Anipreps+%2Fimport+nibabel+as+nib%2F&type=code): 273 files
* [`/import nibabel as nib/`](https://github.com/search?q=org%3Anipreps+%2Fimport+nibabel+as+nib%2F&type=code): 12 files

If we want to consolidate to one spelling, it's definitely the lower effort option to adopt `nb` universally, and we might take this difference as a rough vote. That said, I'm okay with proposing a change to the group and adopting it universally.

Whatever the outcome, I think it would help to add [lint.flake8-import-conventions.extend-aliases](https://docs.astral.sh/ruff/settings/#lint_flake8-import-conventions_extend-aliases) and enable [ICN001](https://docs.astral.sh/ruff/rules/unconventional-import-alias/) to our `ruff` configs to enforce and autofix this.

jhlegarreta:
> If we want to consolidate to one spelling, it's definitely the lower effort option to adopt nb universally, and we might take this difference as a rough vote. That said, I'm okay with proposing a change to the group and adopting it universally.

Changed to `nb`.

> Whatever the outcome, I think it would help to add [lint.flake8-import-conventions.extend-aliases](https://docs.astral.sh/ruff/settings/#lint_flake8-import-conventions_extend-aliases) and enable [ICN001](https://docs.astral.sh/ruff/rules/unconventional-import-alias/) to our ruff configs to enforce and autofix this.

PR #46.

PR #50, opened 2025-01-16
ENH: Provide masks to models in tests

jhlegarreta:
- ENH: Provide masks to models in tests
- BUG: Fill mask parameter for trivial model in model factory

### DIFF
--- a/src/nifreeze/model/base.py
+++ b/src/nifreeze/model/base.py
@@ -50,7 +50,9 @@ class ModelFactory:
 
         """
         if model.lower() in ("s0", "b0"):
-            return TrivialModel(predicted=kwargs.pop("S0"), gtab=kwargs.pop("gtab"))
+            return TrivialModel(
+                mask=kwargs.pop("mask"), predicted=kwargs.pop("S0"), gtab=kwargs.pop("gtab")
+            )
 
         if model.lower() in ("avgdwi", "averagedwi", "meandwi"):
             from nifreeze.model.dmri import AverageDWIModel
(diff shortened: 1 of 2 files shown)


jhlegarreta:
There is other places where masks would be needed, e.g.
https://github.com/nipreps/nifreeze/blob/main/test/test_model.py#L156
https://github.com/nipreps/nifreeze/blob/main/test/test_integration.py#L65

Thus the draft PR.

The current test dataset does not contain a brainmask. A solution could be to compute a brainmask on the b0, include it in the HDF5 file, and push it.

jhlegarreta:
> The current test dataset does not contain a brainmask. A solution could be to compute a brainmask on the b0, include it in the HDF5 file, and push it.

Went through the HDF5 test dataset to see what it looks like and see what an Otsu on the bzero would give: wrote the values to a NIfTI to check. The b0 looks like an actual brain image; the DWI data shows nothing on three different image viewers, despite having some nonzero values. Maybe I'm wrong, but here is a gist of the script I used:
https://gist.github.com/jhlegarreta/6e558ebb732933991c3f09f2a46d0581

Here is the gist to read the HDF5 and to add the brainmask content:
https://gist.github.com/jhlegarreta/8527d44b7d2139d2e510b60e776d6e01

And I can share the file itself through GIN.

jhlegarreta:
> The current test dataset does not contain a brainmask. A solution could be to compute a brainmask on the b0, include it in the HDF5 file, and push it.

OK, I see now that my local testing HDF5 file did not have one, but the one in `g-node` does have one:
https://gin.g-node.org/nipreps-data/tests-nifreeze/commit/3db0a9e161d269ca5ce1b64279f56469c142a47b

So this ready to be merged.

oesteban (approved):

PR #55, opened 2025-01-19
MAINT: Add benchmarking files

jhlegarreta:
Add benchmarking files so that `nifreeze` can be benchmarked using `asv`:
- Add the actual files that allow to benchmark different `nifreeze` capabilities.
- Add a new `benchmark` optional dependencies section to `pyproject.toml`.
- Add the `asv` configuration file.
- Add a `README.rst` file to explain how to run the benchmarking.
- Add a GitHub Actions workflow file to run the benchmarks for every PR.

### DIFF
--- a/benchmarks/README.rst
+++ b/benchmarks/README.rst
@@ -1,0 +1,41 @@
+..  -*- rst -*-
+
+===================
+NiFreeze benchmarks
+===================
+Benchmarking NiFreeze with Airspeed Velocity.
+
+Usage
+-----
+Airspeed Velocity manages building and Python environments by itself,
+unless told otherwise.
+To run the benchmarks, you do not need to install
+a development version of *NiFreeze* on your current
+*Python* environment.
+
+To run all benchmarks for the latest commit, navigate to *NiFreeze*'s root
+``benchmarks`` directory and execute::
+
+    asv run
+
+For testing benchmarks locally, it may be better to run these without
+replications::
+
+    export REGEXP="bench.*Ufunc"
+    asv run --dry-run --show-stderr --python=same --quick -b $REGEXP
+
+All of the commands above display the results in plain text in the console,
+and the results are not saved for comparison with future commits.
+For greater control, a graphical view, and to have results saved for future
+comparisons, you can run ASV as follows to record results and generate
+the HTML reports::
+
+    asv run --skip-existing-commits --steps 10 ALL
+    asv publish
+    asv preview
+
+More on how to use ``asv`` can be found in the `ASV documentation`_.
+Command-line help is available as usual via ``asv --help`` and
+``asv run --help``.
+
+.. _ASV documentation: https://asv.readthedocs.io/
--- a/benchmarks/asv.conf.json
+++ b/benchmarks/asv.conf.json
@@ -1,0 +1,98 @@
+{
+    // The version of the config file format.  Do not change, unless
+    // you know what you are doing.
+    "version": 1,
+
+    // The name of the project being benchmarked
+    "project": "nifreeze",
+
+    // The project's homepage
+    "project_url": "https://www.nipreps.org/nifreeze/",
+
+    // The URL or local path of the source code repository for the
+    // project being benchmarked
+    "repo": "..",
+
+    // List of branches to benchmark. If not provided, defaults to "master"
+    // (for git) or "tip" (for mercurial).
+    "branches": ["HEAD"],
+
+    "build_command": [
+        "python -m build --wheel -o {build_cache_dir} {build_dir}"
+    ],
+
+    // The DVCS being used.  If not set, it will be automatically
+    // determined from "repo" by looking at the protocol in the URL
+    // (if remote), or by looking for special directories, such as
+    // ".git" (if local).
+    "dvcs": "git",
+
+    // The tool to use to create environments.  May be "conda",
+    // "virtualenv" or other value depending on the plugins in use.
+    // If missing or the empty string, the tool will be automatically
+    // determined by looking for tools on the PATH environment
+    // variable.
+    "environment_type": "virtualenv",
+
+    // the base URL to show a commit for the project.
+    "show_commit_url": "https://github.com/nipreps/nifreeze/commit/",
+
+    // The Pythons you'd like to test against.  If not provided, defaults
+    // to the current version of Python used to run `asv`.
+    // "pythons": ["3.12"],
+
+    // The matrix of dependencies to test.  Each key is the name of a
+    // package (in PyPI) and the values are version numbers.  An empty
+    // list indicates to just test against the default (latest)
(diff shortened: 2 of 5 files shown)


jhlegarreta:
Thought this would be relevant for when we will be needing to measure/benchmark `nifreeze`. This is a shy attempt to it. A few thoughts:
- Maybe a smaller testing/benchmarking dataset should be used.
- This PR is by no means meant to be comprehensive; it is meant to be an initial step: further abstraction and sophistication so that we avoid boilerplate code, etc. (e.g. the `scikit-learn` way could serve as an inspiration) are left for a separate PR.
- The `asv.conf.json` file is left at the project root. Let me know if we want to move this to the `benchmarks` folder.
- Eventually we will be willing to compare across different commits/over time so that we avoid introducing changes that are detrimental to performance.
- May need to pin dependencies, which will also require maintenance, since at some point we will want to measure the performance with modern, faster versions of packages.

An issue could be opened to keep track of the relevant comments above.

benchmarks/README.rst
```diff
@@ -0,0 +1,40 @@
+..  -*- rst -*-
+
+===================
+NiFreeze benchmarks
+===================
+
```
oesteban, line 6: Rst does not leave empty lines after titles:
```suggestion
```

benchmarks/README.rst
```diff
@@ -0,0 +1,40 @@
...
+Benchmarking NiFreeze with Airspeed Velocity.
+
+Usage
+-----
+
```
oesteban, line 11:
```suggestion
```

benchmarks/README.rst
```diff
@@ -0,0 +1,40 @@
...
+Airspeed Velocity manages building and Python environments by itself,
+unless told otherwise. To run the benchmarks, you do not need to
+install a development version of NiFreeze to your current Python environment.
```
oesteban, line 14: Semantic line breaks:
```suggestion
unless told otherwise.
To run the benchmarks, you do not need to install
a development version of *NiFreeze* on your current
*Python* environment.
```

benchmarks/README.rst
```diff
@@ -0,0 +1,40 @@
...
+
+To run all benchmarks for the latest commit, navigate to the root NiFreeze
```
oesteban, line 16:
```suggestion
To run all benchmarks for the latest commit, navigate to *NiFreeze*'s root
```

benchmarks/README.rst
```diff
@@ -0,0 +1,40 @@
...
+For testing benchmarks locally, it may be better to run these without
+replications::
+
+    export REGEXP="bench.*Ufunc"
+    asv run --dry-run --show-stderr --python=same --quick -b $REGEXP
+
+All of the commands above display the results in plain text in the console,
+and the results are not saved for comparison with future commits. For
+greater control, a graphical view, and to have results saved for future
+comparison you can run ASV commands (record results and generate HTML):::
```
oesteban, line 30:
```suggestion
and the results are not saved for comparison with future commits.
For greater control, a graphical view, and to have results saved for future
comparisons, you can run ASV as follows to record results and generate
the HTML reports::
```

benchmarks/README.rst
```diff
@@ -0,0 +1,40 @@
...
+
+    asv run --skip-existing-commits --steps 10 ALL
+    asv publish
+    asv preview
+
+More on how to use ``asv`` can be found in `ASV documentation`_
```
oesteban, line 36:
```suggestion
More on how to use ``asv`` can be found in the `ASV documentation`_.
```

benchmarks/asv.conf.json
```diff
@@ -0,0 +1,98 @@
...
+    // with results. If the commit is `null`, regression detection is
+    // skipped for the matching benchmark.
+    //
+    // "regressions_first_commits": {
+    //    "some_benchmark": "352cdf",  // Consider regressions only after this commit
+    //    "another_benchmark": null,   // Skip regression detection altogether
+    // }
+
+    // Maximum time in seconds that a benchmark is allowed to run before it is terminated.
+    "default_benchmark_timeout": 240,
```
oesteban, line 97:
```suggestion
    "default_benchmark_timeout": 240
```

oesteban (commented):
I'm not fully sure what's ASV about, but it looks useful :) Happy to see forward-looking initiatives @jhlegarreta.

benchmarks/README.rst
jhlegarreta: We should add an rst linter: different people will disagree on this.

benchmarks/asv.conf.json
jhlegarreta: Sorry, when creating an initial, empty `asv` config file, the tool adds commas after every value, including the last item. You can check this locally. So I am only following the pattern across the rest of the file here.

oesteban: I don't know what they use for parsing the JSON, but any standard parser will crash with this one. Assuming their parser does not crash with valid JSON, I'd go with valid JSON. (In addition, this error trips highlighters such as GitHub's)

jhlegarreta: Other major tools like `NumPy` or `scikit-learn` seem to keep the commas, but OK, I'd rather use an automatic tool to lint the JSON file then. Can you please propose one?

oesteban: I don't think we use sufficient JSONs as to set up a linter, it's just that if we catch that ASV's config initialization is lazy we should not inherit a malformed JSON object. I honestly doubt we are going to keep editing this file.

A plausible scenario is that the start using a proper parser in some future version of avs and then our tests break because this JSON was malformed all the way through. It's just about keeping things tidy.

benchmarks/README.rst
jhlegarreta: > A linter cannot deal with semantics...

OK, valid point, but my point is to reduce reviewing and maintenance burden, and avoiding inconsistencies across all NiPreps. It is even hard to enforce this within a single repository. I'd rather adopt a linter and content ourselves with however that linter breaks lines, even if they are not semantic breaks. Requires effort to get used to probably, but involves less maintenance.

https://pypi.org/project/restructuredtext-lint/
https://pypi.org/project/sphinx-lint/

Maybe there are better ones as well.

> The question is whether you or other maintainer disagrees :)

:white_flag: if I am the only one disagreeing? Topic for tech call?

benchmarks/asv.conf.json
jhlegarreta: OK, removed. Force pushed.

oesteban: It looks like they are planning to move to toml. I just realized that the comments are also disallowed in JSON (this could be read with something like hJSON). Let's move on and how they adopt toml soon.

oesteban (approved):

PR #56, opened 2025-01-19
STYLE: Increase consistency in average DWI model class name

jhlegarreta:
- STYLE: Increase consistency in average DWI model class name
- DOC: Honor the purpose of the average DWI model in class docstring


jhlegarreta:
Since there is no release yet, API is not being broken by this.

oesteban (approved):
I believe we will end up having two models (one specific for DWI and one general), but I agree this is necessary. Merging in.

PR #57, opened 2025-01-20
ENH: Update LOVO splitter to new dataset indexed access

oesteban:
Leverage the new ``__getitem__`` interface, which provides uniform indexed access across modalities.

Requires: #52.
Related-To: #19.
Resolves: #20.


jhlegarreta (approved):
Thanks for doing this :100:.

Issue #21, opened 2024-05-16
Motion estimator class (estimator.py) could be made more generic to accommodate other data types (e.g. PET)


mnoergaard:
### What would you like to see added in this software?

Currently, the EddyMotionEstimator is targeted DWI data, however, several of the code can be re-used to other data types e.g. PET. Alternatively, we can make a PETMotionEstimator (branch enh/pet_model) that can estimate the motion. 

### Do you have any interest in helping implement the feature?

Yes

Issue #12, opened 2022-12-15
Using a sequence of "b0" models doesn't currently work


arokem:
Using: 

```
models = ("b0", "b0")
...

em_affines = estimator.fit(
        dwdata=dwi_motion, align_kwargs=None, models=models, seed=None
    )

```

results in an error: 

```
No such file or directory: '/home/tmgomez/eddymotion/src/eddymotion/config/dwi-to-b0_level1.json'
```

Is there a particular way to pass a sequence of models that avoids this? Or maybe we are not intended to use the "b0" model more than once?

arokem:
Yeah - I think that would be a great solution for this and very flexible: IIUc, this can accommodate multiple layers of recursion, and can potentially accommodate other types of estimators as "preprocessing" steps if we want to implement those down the line. I guess we could also implement this by replacing `input` instead of the current `data` argument, and if it's another `Estimator` class instance, you trigger that one first?

arokem:
Yep - precisely! WDYT?

Issue #16, opened 2024-04-16
Should parallelization be handled upstream?


arokem:
This is a suggestion to simplify this: 

https://github.com/nipreps/eddymotion/blob/b1f70cc67417ac2fab43ccf5196b9b26397b2326/src/eddymotion/model/base.py#L160

We are in the midst of some work to parallelize model fitting in DIPY itself: https://github.com/dipy/dipy/pull/2593, which would benefit all downstream uses of these models, including here. I wonder whether we should aim to use these methods as implemented there (e.g., by passing through parallelization kwargs to DIPY's multi-voxel parallelization methods), instead of implementing that separately here.

PR #62, opened 2025-01-21
ENH: Implicitly-pipelined and modality-agnostic ``Estimator``

oesteban:
Changes our current implementation of the estimator with a new architecture that allows stacking
(https://github.com/nipreps/nifreeze/issues/12#issuecomment-2598471780):

```Python
estimator_level1 = Estimator(model="b0", ...)    # e.g., 6 dof registration
estimator_level2 = Estimator(model="b0", input=estimator_level1, ...) # e.g., 9 dof registration

estimator_level2.fit(dataset_object)  # Checks "input" and it's another estimator, so runs that first
```

This allow for adding *filters* (to be implemented):

``` Python
downsample = Filter(...)  # e.g., downsampling filter
estimator_level1 = Estimator(model="b0", input=downsample, ...)    # e.g., 6 dof registration
estimator_level2 = Estimator(model="b0", input=estimator_level1, ...) # e.g., 9 dof registration

estimator_level2.fit(dataset_object)  # Checks "input" and it's another estimator, so runs that first
```

In this case, the filtered dataset only feeds ``estimator_level1``. The second level will work on a full dataset.
If you want to interleave another downsampling filter:

``` Python
downsample1 = Filter(...)  # e.g., downsampling filter 1
estimator_level1 = Estimator(model="b0", input=downsample, ...)    # e.g., 6 dof registration
downsample2 = Filter(input=estimator_level1, ...)  # e.g., downsampling filter 2
estimator_level2 = Estimator(model="b0", input=downsaple2, ...) # e.g., 9 dof registration

estimator_level2.fit(dataset_object)  # Checks "input" and it's another estimator, so runs that first
```

Resolves: #12.
Resolves: #21.

### DIFF
--- a/docs/notebooks/bold_realignment.ipynb
+++ b/docs/notebooks/bold_realignment.ipynb
@@ -337,7 +337,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from nifreeze.model.base import AverageModel\n",
+    "from nifreeze.model.base import ExpectationModel\n",
     "from nifreeze.utils.iterators import random_iterator"
    ]
   },
@@ -358,7 +358,7 @@
     "        t_mask[t] = True\n",
     "\n",
     "        # Fit and predict using the model\n",
-    "        model = AverageModel()\n",
+    "        model = ExpectationModel()\n",
     "        model.fit(\n",
     "            data[..., ~t_mask],\n",
     "            stat=\"median\",\n",
@@ -376,7 +376,7 @@
     "            fixedmask_path=brainmask_path,\n",
     "            output_transform_prefix=f\"conversion-{t:02d}\",\n",
     "            num_threads=8,\n",
-    "        )\n",
+    "        ).cmdline\n",
     "\n",
     "        # Run the command\n",
     "        proc = await asyncio.create_subprocess_shell(\n",
--- a/scripts/optimize_registration.py
+++ b/scripts/optimize_registration.py
@@ -133,7 +133,7 @@ async def train_coro(
                 fixedmask_path=brainmask_path,
                 output_transform_prefix=f"conversion-{index:04d}",
                 **align_kwargs,
-            )
+            ).cmdline
 
             tasks.append(
                 ants(
--- a/src/nifreeze/cli/run.py
+++ b/src/nifreeze/cli/run.py
@@ -42,12 +42,17 @@ def main(argv=None) -> None:
     # Open the data with the given file path
     dataset: BaseDataset = BaseDataset.from_filename(args.input_file)
 
-    estimator: Estimator = Estimator()
+    prev_model: Estimator | None = None
+    for model in args.models:
+        estimator: Estimator = Estimator(
+            args.model,
+            prev=prev_model,
+        )
+        prev_model = estimator
 
     _ = estimator.run(
         dataset,
         align_kwargs=args.align_config,
-        models=args.models,
         omp_nthreads=args.nthreads,
         njobs=args.njobs,
         seed=args.seed,
--- a/src/nifreeze/cli/run.py
+++ b/src/nifreeze/cli/run.py
@@ -45,7 +45,7 @@ def main(argv=None) -> None:
     prev_model: Estimator | None = None
     for _model in args.models:
         estimator: Estimator = Estimator(
-            args.model,
+            _model,
             prev=prev_model,
         )
         prev_model = estimator
--- a/src/nifreeze/cli/run.py
+++ b/src/nifreeze/cli/run.py
@@ -25,7 +25,7 @@
 from pathlib import Path
 
 from nifreeze.cli.parser import parse_args
-from nifreeze.data.dmri import DWI
+from nifreeze.data.base import BaseDataset
 from nifreeze.estimator import Estimator
 
 
@@ -40,12 +40,12 @@ def main(argv=None) -> None:
     args = parse_args(argv)
 
     # Open the data with the given file path
-    dwi_dataset: DWI = DWI.from_filename(args.input_file)
+    dataset: BaseDataset = BaseDataset.from_filename(args.input_file)
 
     estimator: Estimator = Estimator()
 
-    _ = estimator.estimate(
-        dwi_dataset,
+    _ = estimator.run(
+        dataset,
         align_kwargs=args.align_config,
         models=args.models,
         omp_nthreads=args.nthreads,
@@ -58,7 +58,7 @@ def main(argv=None) -> None:
     output_path: Path = Path(args.output_dir) / output_filename
 
     # Save the DWI dataset to the output path
-    dwi_dataset.to_filename(output_path)
+    dataset.to_filename(output_path)
 
 
 if __name__ == "__main__":
--- a/src/nifreeze/data/dmri.py
+++ b/src/nifreeze/data/dmri.py
@@ -37,6 +37,30 @@ from nitransforms.linear import Affine
 
 from nifreeze.data.base import BaseDataset, _cmp, _data_repr
 
+DEFAULT_CLIP_PERCENTILE = 75
+"""Upper percentile threshold for intensity clipping."""
+
+DEFAULT_MIN_S0 = 1e-5
+"""Minimum value when considering the :math:`S_{0}` DWI signal."""
+
+DEFAULT_MAX_S0 = 1.0
+"""Maximum value when considering the :math:`S_{0}` DWI signal."""
+
+DEFAULT_LOWB_THRESHOLD = 50
+"""The lower bound for the b-value so that the orientation is considered a DW volume."""
+
+DEFAULT_HIGHB_THRESHOLD = 8000
+"""A b-value cap for DWI data."""
+
+DEFAULT_NUM_BINS = 15
+"""Number of bins to classify b-values."""
+
+DEFAULT_MULTISHELL_BIN_COUNT_THR = 7
+"""Default bin count to consider a multishell scheme."""
+
+DTI_MIN_ORIENTATIONS = 6
+"""Minimum number of nonzero b-values in a DWI dataset."""
+
 
 @attr.s(slots=True)
 class DWI(BaseDataset):
@@ -221,7 +245,7 @@ def load(
     bvec_file: Path | str | None = None,
     bval_file: Path | str | None = None,
     b0_file: Path | str | None = None,
-    b0_thres: float = 50.0,
+    b0_thres: float = DEFAULT_LOWB_THRESHOLD,
 ) -> DWI:
     """
     Load DWI data and construct a DWI object.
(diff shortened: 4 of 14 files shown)


docs/notebooks/bold_realignment.ipynb
```diff
@@ -337,7 +337,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from nifreeze.model.base import AverageModel\n",
+    "from nifreeze.model.base import ExpectationModel\n",
```
oesteban, line 340: Average is too specific about the statistic used. If *Expectation* meets opposition, I'm happy to settle whatever others think is best (or even revert back to Average).

scripts/optimize_registration.py
```diff
@@ -134,7 +134,7 @@ async def train_coro(
                 moving_path,
                 fixedmask_path=brainmask_path,
                 **_kwargs,
-            )
+            ).cmdline
```
oesteban, line 137: The `generate_command` function has been modified to return the nipype interface and that way unify the antsRegistration command line generation in a single function.

src/nifreeze/cli/run.py
```diff
@@ -40,14 +40,19 @@ def main(argv=None) -> None:
     args = parse_args(argv)
 
     # Open the data with the given file path
-    dwi_dataset: DWI = DWI.from_filename(args.input_file)
+    dataset: BaseDataset = BaseDataset.from_filename(args.input_file)
```
oesteban, line 43: Generalize beyond DWI. In the end, we most likely will not expose the dataset HDF5 and accept NIfTIs + metadata (e.g., bvecs/bvals) and build the appropriate object. For a future PR.

src/nifreeze/data/dmri.py
```diff
@@ -39,6 +39,30 @@
...
+"""A b-value cap for DWI data."""
+
+DEFAULT_NUM_BINS = 15
+"""Number of bins to classify b-values."""
+
+DEFAULT_MULTISHELL_BIN_COUNT_THR = 7
+"""Default bin count to consider a multishell scheme."""
+
+DTI_MIN_ORIENTATIONS = 6
+"""Minimum number of nonzero b-values in a DWI dataset."""
```
oesteban, line 64: These variables better belong in the data representation object rather than models (perhaps the two first could be returned back to model).

src/nifreeze/data/dmri.py
```diff
@@ -342,3 +366,87 @@ def load(
         dwi_obj.brainmask = np.asanyarray(mask_img.dataobj, dtype=bool)
 
     return dwi_obj
+
+
+def find_shelling_scheme(
```
oesteban, line 371: This felt like a dwi-specific manipulation routine, I moved it here for that reason.

src/nifreeze/data/dmri.py
```diff
@@ -342,3 +366,87 @@ def load(
...
+        scheme = "single-shell"
+    elif nonempty_bins < multishell_nonempty_bin_count_thr:
+        scheme = "multi-shell"
+    else:
+        scheme = "DSI"
+
+    return scheme, bval_groups, bval_estimated
+
+
+def _rasb2dipy(gradient):
```
oesteban, line 433: This one also felt misplaced. And it's not really necessary (see #3)

src/nifreeze/estimator.py
```diff
@@ -22,229 +22,151 @@
...
+from nifreeze.data.base import BaseDataset
+from nifreeze.model.base import BaseModel, ModelFactory
+from nifreeze.registration.ants import (
+    _prepare_registration_data,
+    _run_registration,
+)
 from nifreeze.utils import iterators
 
 
+class Filter:
```
oesteban, line 42: A stub for now.

src/nifreeze/model/base.py
```diff
@@ -78,114 +79,81 @@ class BaseModel:
...
-        return self._is_fitted
-
-    def fit(self, data, **kwargs):
-        """Abstract member signature of fit()."""
-        raise NotImplementedError("Cannot call fit() on a BaseModel instance.")
-
-    def predict(self, *args, **kwargs):
-        """Abstract member signature of predict()."""
-        raise NotImplementedError("Cannot call predict() on a BaseModel instance.")
+    def fit_predict(self, *_, **kwargs):
```
oesteban, line 97: We don't need to expose a fit/predict interface here. We'll fit&predict all the time. For stationary models (meaning, those trained on the full dataset, e.g., to fine tune hyperparameters such as GPR, will need to implement this logic)

src/nifreeze/model/base.py
```diff
@@ -78,114 +79,81 @@ class BaseModel:
...
 
-        super().__init__(**kwargs)
-        self._predicted = predicted
-        self._datashape = predicted.shape
+        super().__init__(dataset, **kwargs)
+        self._predicted = (
+            predicted
+            if predicted is not None
+            # Infer from dataset if not provided at initialization
+            else getattr(dataset, "reference", getattr(dataset, "bzero", None))
```
oesteban, line 118: Reference comes from thinking of, e.g., fMRI's SBRefs.

src/nifreeze/model/base.py
```diff
@@ -78,114 +79,81 @@ class BaseModel:
...
-        "_model",
-        "_mask",
-        "_models",
-        "_datashape",
-        "_is_fitted",
-        "_modelargs",
-    )
+    __slots__ = {
+        "_dataset": "Reference to a :obj:`~nifreeze.data.base.BaseDataset` object.",
+    }
```
oesteban, line 84: For ``__slots__`` to be documented with autodoc you write them as a dict.

src/nifreeze/model/base.py
```diff
@@ -78,114 +79,81 @@ class BaseModel:
...
+        super().__init__(dataset, **kwargs)
+        self._stat = stat
 
-    def fit(self, data, **kwargs):
-        """Calculate the average."""
+    def fit_predict(self, index, *_, **kwargs):
+        """
+        Return the expectation map.
 
-        # Regress out global signal differences
```
oesteban, line 165: If detrending is necessary, the right way to do this is a filter, in this case.

src/nifreeze/model/dmri.py
```diff
@@ -209,21 +159,21 @@ def predict(self, gradient=None, **kwargs):
...
+            retval[brainmask, ...] = predicted
         else:
-            retval = predicted.reshape(self._datashape[:-1])
+            retval = predicted.reshape(self._dataset.shape[:-1])
 
         return retval
 
 
-class AverageDWIModel(BaseDWIModel):
+class AverageDWIModel(ExpectationModel):
```
oesteban, line 171: I held up changing Average here until I got a sense of how you all receive "Expectation". The final merge should have consistent naming.

oesteban:
This refactor opens the door to easily address #16 and refrain from having all models derive from a parallelized base.

src/nifreeze/estimator.py
```diff
@@ -22,229 +22,151 @@
...
+        dataset : :obj:`~nifreeze.data.base.BaseDataset`
+            The input dataset this estimator operates on.
+
+        Returns
+        -------
+        :obj:`~nifreeze.estimator.Estimator`
+            The estimator, after fitting.
+
+        """
+        return dataset
```
effigies, line 60:
```suggestion
    def run(self, dataset: BaseDataset, **kwargs) -> BaseDataset:
        """
        Trigger execution of the designated filter.

        Parameters
        ----------
        dataset : :obj:`~nifreeze.data.base.BaseDataset`
            The input dataset this estimator operates on.

        Returns
        -------
        dataset : :obj:`~nifreeze.data.base.BaseDataset`
            The dataset, after filtering.

        """
        return dataset
```

Really, you want to do something like:

```python
DatasetT = ty.TypeVar('DatasetT', bound=BaseDataset)

class Filter:
    def run(self, dataset: DatasetT, **kwargs) -> DatasetT:
        return dataset
```

This ensures that, if `pet_ds` is a `PETDataset`, then a `Filter().run(pet_ds)` will be understood to return a `PETDataset`, and likewise any other dataset type.

src/nifreeze/estimator.py
```diff
@@ -22,229 +22,151 @@
...
-        models=("b0",),
-        omp_nthreads=None,
-        n_jobs=None,
+    __slots__ = ("_model", "_strategy", "_prev", "_model_kwargs", "_align_kwargs")
+
+    def __init__(
+        self,
+        model: BaseModel | str,
+        strategy: str = "random",
+        prev: Self | None = None,
```
effigies, line 72: I think this means that `EstimatorA(prev=EstimatorA())` works but `EstimatorB(prev=EstimatorB())` won't. If any subclass should work, then:

```suggestion
        prev: Estimator | Filter | None = None,
```

Also adding `Filter`, since that seems a design goal.

src/nifreeze/estimator.py
```diff
@@ -22,229 +22,151 @@
...
     ):
-        r"""
-        Estimate head-motion and Eddy currents.
+        self._model = model
+        self._prev = prev
+        self._strategy = strategy
+        self._model_kwargs = model_kwargs or {}
+        self._align_kwargs = kwargs or {}
+
+    def run(self, dataset: BaseDataset, **kwargs):
```
effigies, line 82:
```suggestion
    def run(self, dataset: BaseDataset, **kwargs) -> Self:
```

src/nifreeze/model/base.py
effigies: Why document a private variable?

src/nifreeze/model/base.py
```diff
@@ -84,111 +83,78 @@ class BaseModel:
...
-        return self._is_fitted
-
-    def fit(self, data, **kwargs):
-        """Abstract member signature of fit()."""
-        raise NotImplementedError("Cannot call fit() on a BaseModel instance.")
-
-    def predict(self, *args, **kwargs):
-        """Abstract member signature of predict()."""
-        raise NotImplementedError("Cannot call predict() on a BaseModel instance.")
+    def fit_predict(self, *_, **kwargs):
```
effigies, line 98: You're specifically permitting arbitrary positional args that will be thrown away? This feels like a strange API. You can disable them entirely with:

```suggestion
    def fit_predict(self, *, **kwargs):
```

That feels like less of a footgun.

RE the NotImplementedError, is this a method that will always be implemented by subclasses? If so, then marking it an `@abstractmethod` might be helpful. Given that you have a `TrivialModel` below, there's no loss to not being able to instantiate a `BaseModel()`.

If it's not always going to exist, you could remove this here, and create a Protocol that can be used to match only `BaseModel`s that implement the method:

```python
class FitPredictModel(ty.Protocol):
    def fit_predict(self, **kwargs): ...
```

Also, the current types are `Any`. Do you know enough to say something like:

```suggestion
    def fit_predict(self, *, **kwargs) -> np.ndarray:
```

effigies (commented):
Some comments. As noted previously, I haven't used this, so I don't have a stake in the API.

src/nifreeze/model/base.py
oesteban: I just left private variable comments on the DWI base class because that'd make it easier to implement models in the future.

oesteban:
Thanks for the useful feedback @effigies :)

I think this is getting over the finish line. This can work as the initial implementation of this approach. We will need to iterate over this PR in the future.

I also think we have almost removed the dMRI bias for the most part. By passing the dataset structure to models, they can query the data for the adequate metadata without exposing it into the estimator object, which should be general.

There's lots to polish, but this PR will facilitate future work, I hope.

jhlegarreta:
Sorry for my late review/reply here. Thanks for doing this.

As for https://github.com/nipreps/nifreeze/pull/62#discussion_r1926999559, does `StatisticalMoment` or `Moment` fit better than `Expectation`?

PR #64, opened 2025-01-23
sty: Address ruff complaints

jhlegarreta:


jhlegarreta:
Can be merged immediately. Making PR #28 more lightweight. I am working on cleaning a little bit some commits to make it easier to review.

effigies:
The removed import doesn't make sense without replacing the nb.load call.

jhlegarreta:
> The removed import doesn't make sense without replacing the nb.load call.

:eagle: :100: removed in the corresponding commit in PR#28 Thanks.

PR #68, opened 2025-01-25
ENH: Add type hints across miscellaneous methods

jhlegarreta:
- ENH: Add type hints across miscellaneous methods
- STYLE: Prefer specifying b-values as integer types
- STYLE: Prefer building an array with `None` to fill `S0`

### DIFF
--- a/src/nifreeze/data/dmri.py
+++ b/src/nifreeze/data/dmri.py
@@ -373,7 +373,7 @@ def find_shelling_scheme(
     bvals: np.ndarray,
     num_bins: int = DEFAULT_NUM_BINS,
     multishell_nonempty_bin_count_thr: int = DEFAULT_MULTISHELL_BIN_COUNT_THR,
-    bval_cap: int = DEFAULT_HIGHB_THRESHOLD,
+    bval_cap: float = DEFAULT_HIGHB_THRESHOLD,
 ) -> tuple[str, list[npt.NDArray[np.floating]], list[np.floating]]:
     """
     Find the shelling scheme on the given b-values.
@@ -391,7 +391,7 @@ def find_shelling_scheme(
         Number of bins.
     multishell_nonempty_bin_count_thr : :obj:`int`, optional
         Bin count to consider a multi-shell scheme.
-    bval_cap : :obj:`int`, optional
+    bval_cap : :obj:`float`, optional
         Maximum b-value to be considered in a multi-shell scheme.
 
     Returns
--- a/src/nifreeze/data/dmri.py
+++ b/src/nifreeze/data/dmri.py
@@ -33,6 +33,7 @@ import attr
 import h5py
 import nibabel as nb
 import numpy as np
+import numpy.typing as npt
 from nibabel.spatialimages import SpatialImage
 from nitransforms.linear import Affine
 
@@ -369,11 +370,11 @@ def load(
 
 
 def find_shelling_scheme(
-    bvals,
-    num_bins=DEFAULT_NUM_BINS,
-    multishell_nonempty_bin_count_thr=DEFAULT_MULTISHELL_BIN_COUNT_THR,
-    bval_cap=DEFAULT_HIGHB_THRESHOLD,
-):
+    bvals: np.ndarray,
+    num_bins: int = DEFAULT_NUM_BINS,
+    multishell_nonempty_bin_count_thr: int = DEFAULT_MULTISHELL_BIN_COUNT_THR,
+    bval_cap: int = DEFAULT_HIGHB_THRESHOLD,
+) -> tuple[str, list[npt.NDArray[np.floating]], list[np.floating]]:
     """
     Find the shelling scheme on the given b-values.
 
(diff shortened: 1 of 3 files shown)


src/nifreeze/model/dmri.py
```diff
@@ -173,7 +175,15 @@ class AverageDWIModel(ExpectationModel):
 
     __slots__ = ("_th_low", "_th_high", "_detrend")
 
-    def __init__(self, dataset, stat="median", th_low=100, th_high=100, detrend=False, **kwargs):
+    def __init__(
+        self,
+        dataset: DWI,
+        stat: str = "median",
+        th_low: numbers.Number = 100,
+        th_high: numbers.Number = 100,
```
effigies, line 183: We should not bother with `numbers`: https://github.com/python/mypy/issues/3186

```suggestion
        th_low: float = 100,
        th_high: float = 100,
```

src/nifreeze/model/dmri.py
```diff
@@ -117,7 +119,7 @@ def _fit(self, index, n_jobs=None, **kwargs):
         self._model = None  # Preempt further actions on the model
         return n_jobs
 
-    def fit_predict(self, index, **kwargs):
+    def fit_predict(self, index: int, **kwargs):
```
effigies, line 122: It looks like you can change L145 to:

```diff
-            S0 = np.array_split(S0, n_models) if S0 is not None else [None] * n_models
+            S0 = np.array_split(S0, n_models) if S0 is not None else np.full(n_models, None)
```

src/nifreeze/data/dmri.py
```diff
@@ -369,10 +369,10 @@ def load(
...
 def find_shelling_scheme(
-    bvals,
-    num_bins=DEFAULT_NUM_BINS,
-    multishell_nonempty_bin_count_thr=DEFAULT_MULTISHELL_BIN_COUNT_THR,
-    bval_cap=DEFAULT_HIGHB_THRESHOLD,
+    bvals: np.ndarray,
+    num_bins: int = DEFAULT_NUM_BINS,
+    multishell_nonempty_bin_count_thr: int = DEFAULT_MULTISHELL_BIN_COUNT_THR,
+    bval_cap: int = DEFAULT_HIGHB_THRESHOLD,
 ):
```
effigies, line 376: It would be good to specify return types, too. Looks (without checking) like:

```suggestion
) -> tuple[str, list[npt.NDArray[np.floating]], list[np.floating]]:
```

You'll need to `import numpy.typing as npt`.

src/nifreeze/model/dmri.py
jhlegarreta: Done. Force pushed.

src/nifreeze/model/dmri.py
```diff
@@ -173,16 +174,24 @@ class AverageDWIModel(ExpectationModel):
...
+        detrend: bool = False,
+        **kwargs,
+    ):
         r"""
         Implement object initialization.
 
         Parameters
         ----------
-        th_low : :obj:`numbers.Number`
+        th_low : :obj:`int`
```
oesteban, line 191: I don't think we want to limit this to only integers, float would be just fine.

src/nifreeze/model/dmri.py
```diff
@@ -173,16 +174,24 @@ class AverageDWIModel(ExpectationModel):
 
     __slots__ = ("_th_low", "_th_high", "_detrend")
 
-    def __init__(self, dataset, stat="median", th_low=100, th_high=100, detrend=False, **kwargs):
+    def __init__(
+        self,
+        dataset: DWI,
+        stat: str = "median",
+        th_low: int = 100,
+        th_high: int = 100,
```
oesteban, line 182:
```suggestion
        th_low: float = 100.,
        th_high: float = 100.,
```

oesteban (commented):
Minor nit pick

src/nifreeze/model/dmri.py
jhlegarreta: Not sure I've ever seen a non-integer b-value, but will change it to `float`.

We may want to reduce the precision here:
https://github.com/nipreps/nifreeze/pull/72/files#diff-774e30bff24e702cda5516cbc6bbbcae1af21b710e2b0ad22cc15741fbed9afeR250

6 decimal places seems a lot, and makes reading the b-val file difficult for a human reader. Maybe fmt="%.1f" or fmt="%.2f" at most?

oesteban: This is not expressing b-values, but excursions from a central b-value. I agree there's probably no difference between integer and float here, but writing float ensures no hidden rounding errors happen.

Re: decimal places. Yes, that doesn't look great. Perhaps we should add an argument for the decimal places and set it to zero for the default.

oesteban (approved):

src/nifreeze/model/dmri.py
jhlegarreta: > Re: decimal places. Yes, that doesn't look great. Perhaps we should add an argument for the decimal places and set it to zero for the default.

Commit 138573f. Resolving.

Issue #63, opened 2025-01-22
Move detrending from ``AverageModel`` to its own ``Filter``


oesteban:
In the context of #62, I'm simplifying the ``AverageModel`` (renamed as ``ExpectationModel`` to better capture that the statistic can be changed) and removing this code:

```Python
        # Regress out global signal differences
        if kwargs.pop("equalize", False):
            data = data.copy().astype("float32")
            reshaped_data = (
                data.reshape((-1, data.shape[-1])) if self._mask is None else data[self._mask]
            )
            p5 = np.percentile(reshaped_data, 5.0, axis=0)
            p95 = np.percentile(reshaped_data, 95.0, axis=0) - p5
            data = (data - p5) * p95.mean() / p95 + p5.mean()
```

This code was there to "detrend" global signals. It should be a filter and not part of the model.

PR #71, opened 2025-01-25
REF: Refactor filters into `filtering`

jhlegarreta:
- FIX: Fix the 4D data generator fixture name
- FIX: Use all provided values in random choice generator fixture
- REF: Refactor filters into `filtering`

Related to #63.

### DIFF
--- a/src/nifreeze/data/filtering.py
+++ b/src/nifreeze/data/filtering.py
@@ -28,8 +28,12 @@ import numpy as np
 from scipy.ndimage import median_filter
 from skimage.morphology import ball
 
+from nifreeze.data.dmri import DEFAULT_CLIP_PERCENTILE
+
 DEFAULT_DTYPE = "int16"
 """The default image's data type."""
+BVAL_ATOL = 100.0
+"""b-value tolerance value."""
 
 
 def advanced_clip(
@@ -96,3 +100,161 @@ def advanced_clip(
         data = np.round(255 * data).astype(dtype)
 
     return data
+
+
+def robust_minmax_normalization(
+    data: np.ndarray,
+    mask: np.ndarray | None = None,
+    p_min: float = 5.0,
+    p_max: float = 95.0,
+    inplace: bool = False,
+) -> np.ndarray | None:
+    r"""Normalize min-max percentiles of each volume to the grand min-max
+    percentiles.
+
+    Robust min/max normalization of the volumes in the dataset following:
+
+    .. math::
+        \text{data}_{\text{normalized}} = \frac{(\text{data} - p_{min}) \cdot p_{\text{mean}}}{p_{\text{range}}} + p_{min}^{\text{mean}}
+
+    where
+
+    .. math::
+        p_{\text{range}} = p_{max} - p_{min}, \quad p_{\text{mean}} = \frac{1}{N} \sum_{i=1}^N p_{\text{range}_i}, \quad p_{min}^{\text{mean}} = \frac{1}{N} \sum_{i=1}^N p_{5_i}
+
+    If a mask is provided, only the data within the mask are considered.
+
+    Parameters
+    ----------
+    data : :obj:`~numpy.ndarray`
+        Data to be normalized.
+    mask : :obj:`~numpy.ndarray`, optional
+        Mask. If provided, only the data within the mask are considered.
+    p_min : :obj:`float`, optional
+        The lower percentile value for normalization.
+    p_max : :obj:`float`, optional
+        The upper percentile value for normalization.
+    inplace : :obj:`bool`, optional
+        If ``False``, the normalization is performed on the original data.
+
+    Returns
+    -------
+    data : :obj:`~numpy.ndarray` or None
+        Normalized data or ``None`` if ``inplace`` is ``True``.
+    """
+
+    normalized = data if inplace else data.copy()
+
+    mask = mask if mask is not None else np.ones(data.shape[-1], dtype=bool)
+    volumes = data[..., mask]
+    reshape_shape = (-1, volumes.shape[-1]) if mask is None else (-1, sum(mask))
+    reshaped_data = volumes.reshape(reshape_shape)
+    p5 = np.percentile(reshaped_data, p_min, axis=0)
+    p95 = np.percentile(reshaped_data, p_max, axis=0) - p5
+    normalized[..., mask] = (volumes - p5) * p95.mean() / p95 + p5.mean()
+
+    if inplace:
+        return None
+
+    return normalized
+
+
+def grand_mean_normalization(
+    data: np.ndarray,
+    mask: np.ndarray | None = None,
+    center: float = DEFAULT_CLIP_PERCENTILE,
+    inplace: bool = False,
+) -> np.ndarray | None:
+    """Robust grand mean normalization.
+
+    Regresses out global signal differences so that data are normalized and
+    centered around a given value.
+
+    If a mask is provided, only the data within the mask are considered.
+
+    Parameters
+    ----------
+    data : :obj:`~numpy.ndarray`
+        Data to be normalized.
+    mask : :obj:`~numpy.ndarray`, optional
+        Mask. If provided, only the data within the mask are considered.
+    center : float, optional
+        Central value around which to normalize the data.
+    inplace : :obj:`bool`, optional
+        If ``False``, the normalization is performed on the original data.
+
+    Returns
+    -------
+    data : :obj:`~numpy.ndarray` or None
+        Normalized data or ``None`` if ``inplace`` is ``True``.
+    """
+
+    normalized = data if inplace else data.copy()
+
+    mask = mask if mask is not None else np.ones(data.shape[-1], dtype=bool)
+    volumes = data[..., mask]
+
+    centers = np.median(volumes, axis=(0, 1, 2))
+    reference = np.percentile(centers[centers >= 1.0], center)
+    centers[centers < 1.0] = reference
+    drift = reference / centers
+    normalized[..., mask] = volumes * drift
+
+    if inplace:
+        return None
+
+    return normalized
+
+
+def dwi_select_shells(
+    gradients: np.ndarray,
+    index: int,
+    atol_low: float | None = None,
+    atol_high: float | None = None,
+) -> np.ndarray:
+    """Select DWI shells around the given index and lower and upper b-value
+    bounds.
+
+    Computes a boolean mask of the DWI shells around the given index with the
+    provided lower and upper bound b-values.
+
+    If ``atol_low`` and ``atol_high`` are both ``None``, the returned shell mask
+    corresponds to the lengths of the diffusion-sensitizing gradients.
+
+    Parameters
+    ----------
+    gradients : :obj:`~numpy.ndarray`
+        Gradients.
+    index : :obj:`int`
+        Index of the shell data.
+    atol_low : :obj:`float`, optional
+        A lower bound for the b-value.
+    atol_high : :obj:`float`, optional
+        An upper bound for the b-value.
+
+    Returns
+    -------
+    shellmask : :obj:`~numpy.ndarray`
+        Shell mask.
+    """
+
+    bvalues = gradients[:, -1]
+    bcenter = bvalues[index]
+
+    shellmask = np.ones(len(bvalues), dtype=bool)
+    shellmask[index] = False  # Drop the held-out index
+
+    if atol_low is None and atol_high is None:
+        return shellmask
+
+    atol_low = 0 if atol_low is None else atol_low
+    atol_high = gradients[:, -1].max() if atol_high is None else atol_high
+
+    # Keep only b-values within the range defined by atol_high and atol_low
+    shellmask[bvalues > (bcenter + atol_high)] = False
+    shellmask[bvalues < (bcenter - atol_low)] = False
+
+    if not shellmask.sum():
+        raise RuntimeError(f"Shell corresponding to index {index} (b={bcenter}) is empty.")
+
+    return shellmask
--- a/src/nifreeze/model/dmri.py
+++ b/src/nifreeze/model/dmri.py
@@ -27,11 +27,8 @@ import numpy as np
 from dipy.core.gradients import gradient_table_from_bvals_bvecs
 from joblib import Parallel, delayed
 
-from nifreeze.data.dmri import (
-    DEFAULT_CLIP_PERCENTILE,
-    DTI_MIN_ORIENTATIONS,
-    DWI,
-)
+from nifreeze.data.dmri import DTI_MIN_ORIENTATIONS, DWI
+from nifreeze.data.filtering import BVAL_ATOL, dwi_select_shells, grand_mean_normalization
 from nifreeze.model.base import BaseModel, ExpectationModel
 
 S0_EPSILON = 1e-6
@@ -215,14 +212,14 @@ class BaseDWIModel(BaseModel):
 class AverageDWIModel(ExpectationModel):
     """A trivial model that returns an average DWI volume."""
 
-    __slots__ = ("_th_low", "_th_high", "_detrend")
+    __slots__ = ("_atol_low", "_atol_high", "_detrend")
 
     def __init__(
         self,
         dataset: DWI,
         stat: str = "median",
-        th_low: float = 100.0,
-        th_high: float = 100.0,
+        atol_low: float = BVAL_ATOL,
+        atol_high: float = BVAL_ATOL,
         detrend: bool = False,
         **kwargs,
     ):
@@ -235,10 +232,10 @@ class AverageDWIModel(ExpectationModel):
             Reference to a DWI object.
         stat : :obj:`str`, optional
             Whether the summary statistic to apply is ``"mean"`` or ``"median"``.
-        th_low : :obj:`float`, optional
+        atol_low : :obj:`float`, optional
             A lower bound for the b-value corresponding to the diffusion weighted images
             that will be averaged.
-        th_high : :obj:`float`, optional
+        atol_low : :obj:`float`, optional
             An upper bound for the b-value corresponding to the diffusion weighted images
             that will be averaged.
         detrend : :obj:`bool`, optional
@@ -249,8 +246,8 @@ class AverageDWIModel(ExpectationModel):
         """
         super().__init__(dataset, stat=stat, **kwargs)
 
-        self._th_low = th_low
-        self._th_high = th_high
+        self._atol_low = atol_low
+        self._atol_high = atol_high
         self._detrend = detrend
 
     def fit_predict(self, index: int | None = None, *_, **kwargs):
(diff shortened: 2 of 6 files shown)


jhlegarreta:
A few comments:
- I think this fits into the philosophy that PR #62 is proposing. 
- 5cec644 created functions, not classes. To be revisited.
- Tests need to be added.
- WIP until the class vs. function and other questions are clarified.

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +101,117 @@ def advanced_clip(
...
+        Detrended data.
+    """
+
+    shelldata = data[..., mask]
+
+    centers = np.median(shelldata, axis=(0, 1, 2))
+    reference = np.percentile(centers[centers >= 1.0], DEFAULT_CLIP_PERCENTILE)
+    centers[centers < 1.0] = reference
+    drift = reference / centers
+    return shelldata * drift
```
jhlegarreta, line 174: Naming for above two functions/filters should be improved.

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +100,114 @@ def advanced_clip(
...
+    shellmask[index] = False
+    shellmask[bvalues > (bcenter + th_high)] = False
+    shellmask[bvalues < (bcenter - th_low)] = False
+
+    if not shellmask.sum():
+        raise RuntimeError(f"Shell corresponding to index {index} (b={bcenter}) is empty.")
+
+    clipped_dataset._dataset = clipped_dataset._dataset.dataobj[..., shellmask]
+
+    return clipped_dataset
```
jhlegarreta, line 213: If we clip the data, maybe the gradients stored in the datasets should also be clipped.

jhlegarreta:
Tests are failing
https://github.com/nipreps/nifreeze/actions/runs/14120081127/job/39558574436?pr=71#step:11:530
```
  >       with pytest.raises(RuntimeError):
  E       Failed: DID NOT RAISE <class 'RuntimeError'>
```

because the exception
https://github.com/nipreps/nifreeze/blob/main/src/nifreeze/model/dmri.py#L226C9-L227

has now been moved to:
https://github.com/nipreps/nifreeze/pull/71/files#diff-2197e776366f74ac177f201dce923642554f122630fabab0b45063f2f6cf1832R209

So this may requires some discussion.

jhlegarreta:
@oesteban I would be grateful if you could review this, even if I have not marked as ready for review. Some comments:
- The tests need to be parameterized.
- I am creating some random data in https://github.com/nipreps/nifreeze/pull/71/files#diff-fce361e6d716b2dd028aac17d1a0472eb1f8ba0ea75ba7b44d12355dae19e08bR57, similar to https://github.com/nipreps/nifreeze/pull/74/files#diff-f311d147ca2f3984f30068c37db4ca0a780e387581219cd6fa50a33e917d12cbR42. So, those methods should eventually be put into a common place (`conftest.py`?) so that we reduce the duplicate code. Thoughts are welcome.

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
         data = np.round(255 * data).astype(dtype)
 
     return data
+
+
+def detrend_data_percentile(data: np.ndarray, mask: np.ndarray | None = None) -> np.ndarray:
```
oesteban, line 107: Perhaps a more representative name and access to the percentiles?
```suggestion
def robust_minmax_normalization(data: np.ndarray, mask: np.ndarray | None = None, pmin: float = 5.0, pmax: float = 95.0) -> np.ndarray:
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+    r"""Detrend data.
```
oesteban, line 108:
```suggestion
    r"""Normalize min-max percentiles of each volume to the grand min-max percentiles.
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+
+    Regresses out global signal differences so that its values are centered around the middle 90%
+    of the data following:
```
oesteban, line 111:
```suggestion
    Robust min/max normalization of the volumes in the dataset following:
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+    Returns
+    -------
+    :obj:`~numpy.ndarray`
+        Detrended data.
+    """
+
+    data = data.copy().astype("float32")
+    reshaped_data = data.reshape((-1, data.shape[-1])) if mask is None else data[mask]
+    p5 = np.percentile(reshaped_data, 5.0, axis=0)
+    p95 = np.percentile(reshaped_data, 95.0, axis=0) - p5
```
oesteban, line 142:
```suggestion
    p5 = np.percentile(reshaped_data, pmin, axis=0)
    p95 = np.percentile(reshaped_data, pmax, axis=0) - p5
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+    return (data - p5) * p95.mean() / p95 + p5.mean()
+
+
+def detrend_dwi_median(data: np.ndarray, mask: np.ndarray | None = None) -> np.ndarray:
```
oesteban, line 146:
```suggestion
def grand_mean_normalization(data: np.ndarray, mask: np.ndarray | None = None, newcenter: float = DEFAULT_CLIP_PERCENTILE) -> np.ndarray:
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+    """Detrend DWI data.
```
oesteban, line 147:
```suggestion
    """Robust grand mean normalization.
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+
+    Regresses out global DWI signal differences so that its standardized and centered around the
+    :data:`src.nifreeze.model.base.DEFAULT_CLIP_PERCENTILE` percentile.
```
oesteban, line 150:
```suggestion
    Regresses out global signal differences so that data are normalized and centered around
    a given percentile.
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+    Returns
+    -------
+    :obj:`~numpy.ndarray`
+        Detrended data.
+    """
+
+    shelldata = data[..., mask]
+
+    centers = np.median(shelldata, axis=(0, 1, 2))
+    reference = np.percentile(centers[centers >= 1.0], DEFAULT_CLIP_PERCENTILE)
```
oesteban, line 170:
```suggestion
    reference = np.percentile(centers[centers >= 1.0], newcenter)
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+    mask : :obj:`~numpy.ndarray`, optional
+        Mask. If provided, only the data within the mask are considered.
+
+    Returns
+    -------
+    :obj:`~numpy.ndarray`
+        Detrended data.
+    """
+
+    shelldata = data[..., mask]
```
oesteban, line 167: This doesn't necessarily need to be DWI.
```suggestion
    volumes = data[..., mask]
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+    th_high : :obj:`float`, optional
+        An upper bound for the b-value.
+
+    Returns
+    -------
+    clipped_dataset : :obj:`~numpy.ndarray`
+        Clipped data.
+    """
+
+    clipped_data = copy.deepcopy(data)
```
oesteban, line 206: The normalizers above, shouldn't the also copy the data?

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+    shelldata = data[..., mask]
+
+    centers = np.median(shelldata, axis=(0, 1, 2))
+    reference = np.percentile(centers[centers >= 1.0], DEFAULT_CLIP_PERCENTILE)
+    centers[centers < 1.0] = reference
+    drift = reference / centers
+    return shelldata * drift
+
+
+def clip_dwi_shell_data(
```
oesteban, line 176:
```suggestion
def dwi_select_shells(
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+    data: np.ndarray,
+    gradients: np.ndarray,
+    index: int,
+    th_low: float = BVAL_THRESHOLD,
+    th_high: float = BVAL_THRESHOLD,
```
oesteban, line 181: Weird, are we not trying to select shells between `th_low` and `th_high`? Also, the specific values should pertain to the caller, not the function---here, I'd just select all shells by default (do nothing).

```suggestion
    th_low: float | None = None,
    th_high: float | None = None,
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+        A lower bound for the b-value.
+    th_high : :obj:`float`, optional
+        An upper bound for the b-value.
+
+    Returns
+    -------
+    clipped_dataset : :obj:`~numpy.ndarray`
+        Clipped data.
+    """
+
```
oesteban, line 205:
```suggestion

    if th_low is None and th_high is None:
        return data
    th_low = 0 if th_low is None else th_low
    th_high = gradients[:, -1].max() if th_high is None else th_high

```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+
+def clip_dwi_shell_data(
+    data: np.ndarray,
+    gradients: np.ndarray,
+    index: int,
+    th_low: float = BVAL_THRESHOLD,
+    th_high: float = BVAL_THRESHOLD,
+) -> np.ndarray:
+    """Clip DWI shell data around the given index and lower and upper b-value bounds.
```
oesteban, line 183: Instead of creating new data, I'd just generate a b-value mask that can then be passed to the above methods.

```suggestion
    """Clip DWI shell data around the given index and lower and upper b-value bounds.
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +102,122 @@ def advanced_clip(
...
+
+    bvalues = gradients[:, -1]
+    bcenter = bvalues[index]
+
+    shellmask = np.ones(clipped_data.shape[-1], dtype=bool)
+
+    # Keep only bvalues within the range defined by th_high and th_low
+    shellmask[index] = False
+    shellmask[bvalues > (bcenter + th_high)] = False
+    shellmask[bvalues < (bcenter - th_low)] = False
```
oesteban, line 216: Return just shellmask?

src/nifreeze/model/dmri.py
```diff
@@ -178,8 +175,8 @@ def __init__(
         self,
         dataset: DWI,
         stat: str = "median",
-        th_low: float = 100.0,
-        th_high: float = 100.0,
+        th_low: float = BVAL_THRESHOLD,
+        th_high: float = BVAL_THRESHOLD,
```
oesteban, line 179: Apologies, this was a really bad name for the parameter:

```suggestion
        atol_low: float = BVAL_ATOL,
        atol_high: float = BVAL_ATOL,
```

oesteban (commented):
There you have some feedback. Thanks for taking on this.

src/nifreeze/data/filtering.py
jhlegarreta: Adopting the naming suggested by Óscar in his review. Thanks Óscar. Resolving.

src/nifreeze/data/filtering.py
jhlegarreta: The selection takes place around the indexed b-value:
https://github.com/nipreps/nifreeze/pull/71/files#diff-2197e776366f74ac177f201dce923642554f122630fabab0b45063f2f6cf1832R214-R215

:+1: as for the rest of the comments. Thanks.

src/nifreeze/data/filtering.py
jhlegarreta: :+1: Will refactor the method after committing the batch of suggestions.

src/nifreeze/data/filtering.py
jhlegarreta: :+1: Agreed. Will do after committing the batch of suggestions.

src/nifreeze/model/dmri.py
```diff
@@ -210,36 +207,26 @@ def __init__(
...
+    def fit_predict(self, index: int, *_, **kwargs):
         """Return the average map."""
 
-        bvalues = self._dataset.gradients[:, -1]
-        bcenter = bvalues[index]
-
-        shellmask = np.ones(len(self._dataset), dtype=bool)
-
-        # Keep only bvalues within the range defined by th_high and th_low
-        shellmask[index] = False
```
jhlegarreta, line 222: @oesteban While writing the test, I have realized that this statement should be removed: we want to keep the value at the `index` position, right?

src/nifreeze/data/filtering.py
jhlegarreta: Done in 641067609fa91c0e77bceae5c3b30d1deac60bed.

src/nifreeze/data/filtering.py
jhlegarreta: Done in 641067609fa91c0e77bceae5c3b30d1deac60bed.

src/nifreeze/data/filtering.py
jhlegarreta: The normalizers above are not changing the data; this was probably a leftover from the version where I was taking a `DWI` data instance. No longer applies as we are now not requiring the data now. Thanks for the review Óscar.

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +100,134 @@ def advanced_clip(
...
+
+    Returns
+    -------
+    :obj:`~numpy.ndarray`
+        Normalized data.
+    """
+
+    volumes = data
+    if mask is not None:
+        volumes = data[..., mask]
```
jhlegarreta, line 174: The check here is necessary otherwise a new, unnecessary dimension is added to the data.

src/nifreeze/model/dmri.py
jhlegarreta: I removed the statement. Consequently, the exception here is not not raised:
https://github.com/nipreps/nifreeze/pull/71/files#diff-2197e776366f74ac177f201dce923642554f122630fabab0b45063f2f6cf1832R230-R231

https://github.com/nipreps/nifreeze/actions/runs/14149369168/job/39640631654?pr=71#step:11:539

So I am wondering about the rationale behind assigning `False` to the index.

jhlegarreta:
Re https://github.com/nipreps/nifreeze/pull/71#issuecomment-2614070194:
- We are happy with methods vs classes.
- Tests have been added in the latest commits.

jhlegarreta:
@oesteban 
- Re https://github.com/nipreps/nifreeze/pull/71#discussion_r2019892104: this is the stepping stone in this PR, and I'd say it is important to find a solution.
- Re https://github.com/nipreps/nifreeze/pull/71#issuecomment-2762934819: tests are parameterized and random data generation functions are added. Transferring them to `conftest.py` can be done in a separate refactoring PR so that we can move forward.

src/nifreeze/model/dmri.py
oesteban: As long as `dwi_select_shells` does this inside (which I imagine it does so), yes, you need to remove this because it's moved into the new function.

oesteban (approved):

src/nifreeze/model/dmri.py
jhlegarreta: It does not, please have a look at the definition https://github.com/nipreps/nifreeze/pull/71/files#diff-2197e776366f74ac177f201dce923642554f122630fabab0b45063f2f6cf1832R183

src/nifreeze/data/filtering.py
oesteban, line 174:
```suggestion
    mask = mask if mask is not None else np.ones(data.shape[-1], dtype=bool)
    volumes = data[..., mask]
```

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +100,134 @@ def advanced_clip(
...
+
+    centers = np.median(volumes, axis=(0, 1, 2))
+    reference = np.percentile(centers[centers >= 1.0], center)
+    centers[centers < 1.0] = reference
+    drift = reference / centers
+    return volumes * drift
```
oesteban, line 180: I don't think this function should filter out masked direction, right?

```suggestion
    data[..., mask] = volumes * drift
    return data
```

src/nifreeze/model/dmri.py
oesteban: > So I am wondering about the rationale behind assigning `False` to the index.

Okay, I just came back to this and now I understand. The idea for this method is: we want to leave one volume out at each iteration, so you want to mark that specific volume for dropping in the returned mask. Does that make sense?

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +100,134 @@ def advanced_clip(
...
+    Returns
+    -------
+    shellmask : :obj:`~numpy.ndarray`
+        Shell mask.
+    """
+
+    bvalues = gradients[:, -1]
+    bcenter = bvalues[index]
+
+    shellmask = np.ones(len(bvalues), dtype=bool)
```
oesteban, line 218:
```suggestion
    shellmask = np.ones(len(bvalues), dtype=bool)
    shellmask[index] = False  # Drop the left-out index
```

jhlegarreta:
BTW, I have run into this: https://github.com/nipreps/nifreeze/blob/main/src/nifreeze/estimator.py#L48

So if the class design that was envisioned in that file should be eventually enforced, the filters in here will need to be morphed into classes. Probably better to do it now if the class-based approach is preferred.

src/nifreeze/data/filtering.py
jhlegarreta: Noting that this was refactored in https://github.com/nipreps/nifreeze/pull/71#discussion_r2083729676.

src/nifreeze/data/filtering.py
```diff
@@ -96,3 +100,135 @@ def advanced_clip(
...
+
+    mask = mask if mask is not None else np.ones(data.shape[-1], dtype=bool)
+    volumes = data[..., mask]
+
+    centers = np.median(volumes, axis=(0, 1, 2))
+    reference = np.percentile(centers[centers >= 1.0], center)
+    centers[centers < 1.0] = reference
+    drift = reference / centers
+    data[..., mask] = volumes * drift
+    return data
```
jhlegarreta, line 180: @oesteban the input argument is being changed here. I usually dislike this, but are we comfortable with it?

src/nifreeze/data/filtering.py
jhlegarreta: There is no clipping of data anymore: the functions returns the shell mask: https://github.com/nipreps/nifreeze/pull/71#discussion_r2019735831; https://github.com/nipreps/nifreeze/pull/71#discussion_r2019735897

src/nifreeze/model/dmri.py
jhlegarreta: The latest forced push d30f522 solves the questions here. Resolving.

jhlegarreta:
@oesteban this is close to being ready to be merged. Please, review [this thread](https://github.com/nipreps/nifreeze/pull/71#discussion_r2105848964) about data copies. Thanks.

src/nifreeze/data/filtering.py
jhlegarreta: Done in 03bc0ac. This is ready to go.

jhlegarreta:
Done with https://github.com/nipreps/nifreeze/pull/85#issuecomment-2925674050: used the fixtures introduced in PR #85. This is ready to be merged. Thanks for the patience.

oesteban (approved):

PR #72, opened 2025-01-25
BUG: Use the appropriate index to extract b-values

jhlegarreta:
- BUG: Use the appropriate index to extract b-values
- STYLE: Add gradient-related properties to DWI data class
- STYLE: Make bvecs/bvals serialized shapes match comments/expectations
- STYLE: Simplify getting the root name for DWI serialization
- DOC: Mark the `insert_b0` parameter as optional
- ENH: Add decimal places for bval/bvec serialization

### DIFF
--- a/src/nifreeze/data/dmri.py
+++ b/src/nifreeze/data/dmri.py
@@ -233,12 +233,8 @@ class DWI(BaseDataset[np.ndarray | None]):
         # Convert filename to a Path object.
         out_root = Path(filename).absolute()
 
-        # Remove .gz if present, then remove .nii if present.
-        # This yields the base stem for writing .bvec / .bval.
-        if out_root.suffix == ".gz":
-            out_root = out_root.with_suffix("")  # remove '.gz'
-        if out_root.suffix == ".nii":
-            out_root = out_root.with_suffix("")  # remove '.nii'
+        # Get the base stem for writing .bvec / .bval.
+        out_root = out_root.parent / out_root.name.replace("".join(out_root.suffixes), "")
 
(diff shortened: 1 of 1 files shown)


jhlegarreta:
A round-trip test is probably missing to prevent issues like the one in 63bdf03. Do not have the bandwidth to do that now.

src/nifreeze/data/dmri.py
```diff
@@ -139,6 +139,14 @@ def from_filename(cls, filename: Path | str) -> DWI:
 
         return cls(**data)
 
+    @property
+    def bvals(self):
+        return self.gradients[:3, ...]
```
oesteban, line 144:
```suggestion
        return self.gradients[-1, ...]
```

jhlegarreta:
Re: https://github.com/nipreps/nifreeze/pull/72#issuecomment-2614125671 There is a round-trip test in https://github.com/nipreps/nifreeze/blob/7322e93bc3eda2ebe5d2036b1f90c0f10596716c/test/test_data_dmri.py#L123

But there is definitely something not working or missing there. BTW, I cannot debug this due to issue #73.

jhlegarreta:
Re https://github.com/nipreps/nifreeze/pull/72#issuecomment-2614484242: issue #79. Since fixing the issue seems to require a bit of work, including maybe some discussion or clarification as I am not convinced by the `insert_b0` option or do not fully understand the rationale of other aspects of the current logic, I would leave fixing #79 for a separate PR.

oesteban (approved):

Issue #73, opened 2025-01-25
Starred expressions in subscriptions and `TypeVarTuple` not available in Python 3.10


jhlegarreta:
### What happened?

When trying to run the `test_data_dmri.py` test suite, Python raises an `ImportError` because it does not like the 
```
class BaseDataset(Generic[*Ts]):
```

syntax.

I am using Python 3.10. According to my IDE:
```
Python version 3.10 does not support starred expressions in subscriptions
```

After removing the problematic `Generic[*Ts]` inheritance, the error is translated to other locations where the `[*Ts]` syntax is used, and once all removed I get
```
ImportError: cannot import name 'TypeVarTuple' from 'typing' (/usr/lib/python3.10/typing.py
```

See the additional information.

These are not being caught because even if we only require [Python >= 3.10](https://github.com/nipreps/nifreeze/blob/7237ecff29902c3e077e06303dc94c0e34ec1ef7/pyproject.toml#L21) , we are only testing with [Python 3.12](https://github.com/nipreps/nifreeze/blob/7237ecff29902c3e077e06303dc94c0e34ec1ef7/.github/workflows/test.yml#L107).

### What command did you use?

```shell
pytest test_data_dmri.py
```

### What version of the software are you running?

`HEAD` at commit 6697c7872cf77e3dd35288118f0294d80d6e7f48

### How are you running this software?

Local installation ("bare-metal")

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

```shell
ImportError while loading conftest 'src/nifreeze/test/conftest.py'.
test/conftest.py:33: in <module>
    from nifreeze.data.dmri import DWI
src/nifreeze/data/dmri.py:39: in <module>
    from nifreeze.data.base import BaseDataset, _cmp, _data_repr
E     File "src/nifreeze/src/nifreeze/data/base.py", line 61
E       class BaseDataset(Generic[*Ts]):
E                         ^^^^^^^^^^^^
E   SyntaxError: invalid syntax. Perhaps you forgot a comma?
```

### Additional information / screenshots

Starred expressions in subscriptions: https://bugs.python.org/issue45586
TypeVarTuple: https://peps.python.org/pep-0646/

effigies:
We should be able to import `TypeVarTuple` from `typing_extensions` and use `Unpack[Ts]` instead of `*Ts`.

PR #74, opened 2025-01-25
ENH: Add a shell data property to `DWI` data class

### LINKED ISSUE #66: Add a ``shells`` property to the ``DWI`` object to access the data split by shell
Instead of repeating the implementation for every DWI model that needs this, write down a property accessor that "bins" b values and returns pairs of (b-value mode, shell data ndarray).

jhlegarreta:
Add a shell data property to `DWI` data class that returns a list of pairs consisting of the estimated b-value and the associated DWI data.

Closes #66.

### DIFF
--- a/test/test_data_dmri.py
+++ b/test/test_data_dmri.py
@@ -25,19 +25,27 @@
 import nibabel as nb
 import numpy as np
 import pytest
+from dipy.io.gradients import read_bvals_bvecs
 
-from nifreeze.data.dmri import find_shelling_scheme, load
+from nifreeze.data.dmri import DWI, find_shelling_scheme, load
 
 
-def _create_dwi_random_dataobj(request):
+def _create_random_gtab_dataobj(request, n_gradients=10, b0s=1):
     rng = request.node.rng
 
-    n_gradients = 10
-    b0s = 1
-    volumes = n_gradients + b0s
-    b0_thres = 50
     bvals = np.hstack([b0s * [0], n_gradients * [1000]])
     bvecs = np.hstack([np.zeros((3, b0s)), rng.random((3, n_gradients))])
+
+    return bvals, bvecs
+
+
+def _create_dwi_random_dataobj(request, bvals, bvecs):
+    rng = request.node.rng
+
+    n_gradients = np.count_nonzero(bvals)
+    b0s = len(bvals) - n_gradients
+    volumes = n_gradients + b0s
+    b0_thres = 50
 
     vol_size = (34, 36, 24)
 
@@ -136,6 +144,8 @@ def test_load(datadir, tmp_path):
 
 def test_equality_operator(tmp_path, request):
     # Create some random data
+    bvals, bvecs = _create_random_gtab_dataobj(request)
+
     (
         dwi_dataobj,
         affine,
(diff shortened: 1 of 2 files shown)


jhlegarreta:
Opened as draft as I cannot debug this locally due to issue #73.

jhlegarreta:
Maybe it is worthwhile thinking about the interplay between this feature and the b-value iterators, and how the whole thing fits into the estimators' pipelining philosophy.

oesteban:
> Maybe it is worthwhile thinking about the interplay between this feature and the b-value iterators, and how the whole thing fits into the estimators' pipelining philosophy.

I agree. In our last *iteration* we made the iterators independent of the object, so now you could go shell by shell and run any available iterator (those without concept of b-value, evidently).

test/test_data_dmri.py
```diff
@@ -182,6 +182,46 @@ def test_equality_operator(tmp_path):
     assert round_trip_dwi_obj == dwi_obj
 
 
+def test_shells(datadir):
+    dwi_h5 = load(datadir / "dwi.h5")
+    num_bins = 3
```
oesteban, line 187: Can we use the bvecs/bvals files we added for testing? They are within the repo (i.e., no need to pull extra data from GIN) and they provide several use-cases ideal for testing this.

jhlegarreta: Not sure if I understand: `shells` is an [instance method](https://github.com/nipreps/nifreeze/pull/74/files#diff-774e30bff24e702cda5516cbc6bbbcae1af21b710e2b0ad22cc15741fbed9afeR245) and thus we need a `DWI` instance. Are you proposing to build a `DWI` instance using random data? I'd say that this would add unnecessary complexity to the test, unless that instance can be generalized and be built to be used by other tests. TBH, I would open an issue if my interpretation is correct and leave that for a separate PR.

jhlegarreta: 390a139 creates random data for the DWI dataobj and uses the bval/bvec data from the repository.

test/test_data_dmri.py
```diff
@@ -25,20 +25,28 @@
...
     bvals = np.hstack([b0s * [0], n_gradients * [1000]])
     bvecs = np.hstack([np.zeros((3, b0s)), rng.random((3, n_gradients))])
 
+    return bvals, bvecs
+
+
+def _create_dwi_random_dataobj(request, bvals, bvecs):
+    rng = request.node.rng
+
+    n_gradients = np.count_nonzero(bvecs)
```
oesteban, line 45: If you use the bvecs for this:
```suggestion
    n_gradients = np.count_nonzero(bvecs, axis=1)
```

oesteban: Otherwise:

```suggestion
    n_gradients = np.count_nonzero(bvals)
```

oesteban (approved):

test/test_data_dmri.py
jhlegarreta: I am resolving this. I am cross-referencing https://github.com/nipreps/nifreeze/pull/71#issuecomment-2762934819. Once I rebase PR #71 on `main` I'll try to figure out how to merge/refactor the random testing data creation methods to a common place.

PR #78, opened 2025-01-26
fix: Test on Python 3.10+, use typing_extensions backports

### LINKED ISSUE #73: Starred expressions in subscriptions and `TypeVarTuple` not available in Python 3.10

effigies:
Closes #73.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -85,6 +85,9 @@ jobs:
     # Unit tests only on Linux/Python 3.12
     runs-on: 'ubuntu-latest'
     needs: ['cache-test-data']
+    strategy:
+      matrix:
+        python-version: ['3.10', '3.11', '3.12']
     steps:
       - uses: actions/checkout@v4
         with:
@@ -93,7 +96,7 @@ jobs:
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
         with:
-          python-version: '3.12'
+          python-version: ${{ matrix.python-version }}
       - name: Install the latest version of uv
         uses: astral-sh/setup-uv@v5
       - name: Install ANTs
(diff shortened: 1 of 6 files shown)


jhlegarreta (approved):
Thanks for doing this @effigies :100:.

PR #80, opened 2025-01-26
chore(ci): Move data cache into own job, key on test data repo HEAD

effigies:
This PR moves data caching into its own job, and creates a new cache iff the remote repository has changed. We don't even need to install datalad if the HEAD hasn't changed.

Just a note that https://gin.g-node.org/nipreps-data/tests-nifreeze having all of its data checked into git instead of git-annex, cloning or updating takes a very long time. We will probably want to change this at some point, and then we'll want to reset the cache version so we don't try to merge unrelated histories.


.github/workflows/test.yml
```diff
@@ -44,40 +94,30 @@ jobs:
...
+          key: data-v1-${{ needs.cache-test-data.outputs.SHA }}
       - name: Install TeX Live
         run: |
           sudo apt-get update
           sudo apt install texlive texlive-latex-extra texlive-fonts-recommended cm-super dvipng
       - name: Install tox
         run: |
-          python -m pip install --upgrade pip
-          python -m pip install tox tox-gh-actions
+          uv tool install --with=tox-uv --with=tox-gh-actions tox
```
effigies, line 120: The switch to `uv tool` instead of `pip` was an experiment in trying to figure out what in the environment caused the tests to fail. I could revert it, but I'll plan to do this in another PR anyway, so I'll leave it unless asked.

oesteban:
> having all of its data checked into git instead of git-annex,

Yes, I've been meaning to change this (and do the right thing) for a long while now. Thanks for making it apparent, otherwise I would have continued ignoring the necessity.

oesteban (approved):
Thanks for this :)

PR #82, opened 2025-01-31
FIX: Revise bzero handling in dMRI data's ``to_nifti()``

### LINKED ISSUE #79: `data.dmri` has mismatches between reading/writting/representation
### What happened?

The DWI reading/saving/representation logic has some bugs.

- The `test_data_mri::test_load` is reading the bvals/bvecs files
https://github.com/nipreps/nifreeze/blob/7322e93bc3eda2ebe5d2036b1f90c0f10596716c/test/test_data_dmri.py#L126-L129
that are being written by the test function
https://github.com/nipreps/nifreeze/blob/7322e93bc3eda2ebe5d2036b1f90c0f10596716c/test/test_data_dmri.py#L112-L113
based on the `grad_table` variable, which is effectively inserting a b0 value to the bvals/bvecs file (having shapes `(1, 103)`, `(3, 103)`).
- However, when `insert_b0` is `True` as it is the case in the test, the `DWI.to_nifti` method:
https://github.com/nipreps/nifreeze/blob/7322e93bc3eda2ebe5d2036b1f90c0f10596716c/src/nifreeze/data/dmri.py#L210-L237
does not insert any null gradient to the bval/bvecs files (having shapes `(1, 102)`, `(3, 102)`). Thus, if we had read those files in the test, the test would be failing due to a shape mismatch error in https://github.com/nipreps/nifreeze/blob/7322e93bc3eda2ebe5d2036b1f90c0f10596716c/src/nifreeze/data/dmri.py#L336

However, there is a deeper bug here:
- If we change the `DWI.to_nifti` method to add a null gradient to the bval/bvecs files:
```
       bvecs = self.bvecs
       bvals = self.bvals

        if not insert_b0:
            # Parent's to_nifti to handle the primary NIfTI export.
            super().to_nifti(filename)
        else:
            data = np.concatenate((self.bzero[..., np.newaxis], self.dataobj), axis=-1)
            bvecs = np.concatenate((np.zeros(3)[:, np.newaxis], bvecs), axis=-1)
            bvals = np.concatenate((np.zeros(1), bvals))
            nii = nb.Nifti1Image(data, self.affine, self.datahdr)
            if self.datahdr is None:
                nii.header.set_xyzt_units("mm")
            nii.to_filename(filename)

        # Convert filename to a Path object.
        out_root = Path(filename).absolute()

        # Get the base stem for writing .bvec / .bval.
        out_root = out_root.parent / out_root.name.replace("".join(out_root.suffixes), "")

        # Construct sidecar file paths.
        bvecs_file = out_root.with_suffix(".bvec")
        bvals_file = out_root.with_suffix(".bval")

        # Save bvecs and bvals to text files
        # Each row of bvecs is one direction (3 rows, N columns).
        np.savetxt(bvecs_file, bvecs, fmt="%.6f")
        np.savetxt(bvals_file, bvals, fmt="%.6f")
```

and if `insert_b0` is `True`, the test will pass. However, it will not pass if `insert_b0` is `False`. None of the cases on lines
 https://github.com/nipreps/nifreeze/blob/7322e93bc3eda2ebe5d2036b1f90c0f10596716c/src/nifreeze/data/dmri.py#L345-L356 are `True` and thus the read `dwi_obj` will have a `None` valuue in ints `bzero` property, whereas the test object `dwi_h5` does have a `bzero` volume, thus `assert np.allclose(dwi_h5.bzero, dwi_from_nifti.bzero)` will fail.

So the logic behind this needs to be reworked. Some thoughts:
- The above shows a clear mismatch of what the object holds, and what is told to write: if we tell to insert a `bzero`, the written file becomes out of sync with its original representation.
- It is confusing to have a `bzero` value (as the original instance read form the `h5` file has), but not to have a 0-valued bval/bvec in the gradients.
- It is equally confusing the fact that the load method accepts an optional b0_file, but no b0_file is being written. Not writing an additional b0 file seems natural, as one expects the b0 file to be saved together with the non-b0 data, but then allowing to read one does not seem natural.
- It is confusing to ask to insert a b0 when writing to some data that already has b0 data: how should this be interpreted and written?
- What happens when a DWI volume has multiple b0 volumes?
- `test_load` should be parameterized with `insert_b0` to `[False, True]` to test all of the above properly.

### What command did you use?

```shell
pytest test_data_mri.py
```

### What version of the software are you running?

`HEAD` (commit 7322e93b) with the necessary fixes in PR #72

### How are you running this software?

Local installation ("bare-metal")

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

```shell

```

X-ref https://github.com/nipreps/nifreeze/pull/72#issuecomment-2614484242.

jhlegarreta:
Add null gradient values to DWI data serialization if appropriate.

When the flag `insert_b0` is `True` the previous implementation prepended the `bzero` attribute data to the `DWI` instance `dataobj`, but it was not adding the corresponding null gradient value to the bval/bvec file pair. This patch set prepends a null gradient to the bval/bvec file pair.

Parameterize the `test_load` testing function to check both cases (i.e. `insert_b0` `False` and `True`).

Add additional checks to ensure that the write/read round-trip works as expected, and that the `bvals` and `bvecs` attributes have the expected values.

Fixes #79.

### DIFF
--- a/src/nifreeze/data/dmri.py
+++ b/src/nifreeze/data/dmri.py
@@ -205,12 +205,16 @@ class DWI(BaseDataset[np.ndarray | None]):
         """
         Write a NIfTI file to disk.
 
+        This method assumes that the ``bzero`` attribute of `~nifreeze.data.dmri.DWI`
+        is not ``None``.
+
         Parameters
         ----------
         filename : :obj:`os.pathlike`
             The output NIfTI file path.
         insert_b0 : :obj:`bool`, optional
-            Insert a :math:`b=0` at the front of the output NIfTI.
+            Insert a :math:`b=0` at the front of the output NIfTI and add the corresponding
+            null gradient value to the output bval/bvec files.
         bvals_dec_places : :obj:`int`, optional
             Decimal places to use when serializing b-values.
         bvecs_dec_places : :obj:`int`, optional
@@ -221,6 +225,7 @@ class DWI(BaseDataset[np.ndarray | None]):
             # Parent's to_nifti to handle the primary NIfTI export.
             super().to_nifti(filename)
         else:
+            # Assume bzero is never None
             data = np.concatenate((self.bzero[..., np.newaxis], self.dataobj), axis=-1)
             nii = nb.Nifti1Image(data, self.affine, self.datahdr)
             if self.datahdr is None:
@@ -237,10 +242,18 @@ class DWI(BaseDataset[np.ndarray | None]):
         bvecs_file = out_root.with_suffix(".bvec")
         bvals_file = out_root.with_suffix(".bval")
 
+        # If inserting a b0 volume is requested, add the corresponding null
+        # gradient value to the bval/bvec pair
+        bvals = self.bvals
+        bvecs = self.bvecs
+        if insert_b0:
+            bvals = np.concatenate((np.zeros(1), bvals))
+            bvecs = np.concatenate((np.zeros(3)[:, np.newaxis], bvecs), axis=-1)
+
         # Save bvecs and bvals to text files
         # Each row of bvecs is one direction (3 rows, N columns).
-        np.savetxt(bvecs_file, self.bvecs, fmt=f"%.{bvecs_dec_places}f")
-        np.savetxt(bvals_file, self.bvals[np.newaxis, :], fmt=f"%.{bvals_dec_places}f")
+        np.savetxt(bvecs_file, bvecs, fmt=f"%.{bvecs_dec_places}f")
+        np.savetxt(bvals_file, bvals[np.newaxis, :], fmt=f"%.{bvals_dec_places}f")
 
     def shells(
         self,
--- a/test/test_data_dmri.py
+++ b/test/test_data_dmri.py
@@ -22,6 +22,8 @@
 #
 """Unit tests exercising the dMRI data structure."""
 
+from pathlib import Path
+
 import nibabel as nb
 import numpy as np
 import pytest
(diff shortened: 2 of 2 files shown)


jhlegarreta:
This only demonstrates the issue; some thinking/discussion is needed to see what convention we follow so that the issue can be addressed properly.

oesteban:
Sure, please book my calendar :)

jhlegarreta:
I think I am done with this: it is ready to be reviewed/merged.

A few comments:

- Do the `dwi-b0_desc-avg.nii.gz`, `dwi-b0_desc-brain.nii.gz` and `dwi-b0_desc-brain_mask.nii.gz` testing data files correspond to what we should expect to be contained in `dwi.h5` testing data file?
  i.e. The below:
   ```
   bzero_brain_nifti = nb.load(datadir / "dwi-b0_desc-brain.nii.gz").get_fdata().astype(np.int16)
   bzero_brain_h5 = np.zeros_like(dwi_h5.bzero)
   bzero_brain_h5[dwi_h5.brainmask] = dwi_h5.bzero[dwi_h5.brainmask]
   assert np.allclose(bzero_brain_nifti, bzero_brain_h5)
   ```
   raises an error. I must be missing something if they should contain the same data.

   https://gin.g-node.org/nipreps-data/tests-nifreeze

- The brainmask is not serialized, so if I were to test that, the round trip would fail. To be added in a separate PR.

oesteban (commented):
I've allowed myself to edit the PR's title because I think it was biasing the solution.

IMHO, #79 partially misrepresents the problem: HDF5 offers built-in serialization (and this is why we are using it, e.g., to access the data from multiple processes/threads without headaches).

I believe the issue is in reality simpler. The object's ``to_nifti()`` has a poor implementation. If we removed ``to_nifti()`` (which honestly, doesn't sound crazy, there's nothing that requires this feature to be part of the data object), then we don't have the problem anymore. It could be a separate function and #79 would read like "``to_nifti()`` doesn't work great".

If that argument can be agreed upon, I've reviewed the PR accordingly. I'm not extracting ``to_nifti()`` from the object (but it possibly makes sense to consider in the future). I'll send a PR against this branch for (IMHO) a more consistent handling of cases when ``bzero`` is not set but ``insert_b0`` is set on.

> I think I am done with this: it is ready to be reviewed/merged.
> 
> A few comments:
> 
> * Do the `dwi-b0_desc-avg.nii.gz`, `dwi-b0_desc-brain.nii.gz` and `dwi-b0_desc-brain_mask.nii.gz` testing data files correspond to what we should expect to be contained in `dwi.h5` testing data file?

I don't remember, TBH. There's no need to stick with them if they do not meet our purposes in testing.

> * The brainmask is not serialized, so if I were to test that, the round trip would fail. To be added in a separate PR.

I'm comfortable with this, I don't think nifreeze should operate on the mask so there's no reason to write it out. It is part of the object because it is necessary in execution time. Recently, I've been thinking we could recommend that this were not a "brain" mask, but rather a conservative "brain parenchyma" mask (i.e., excluding CSF, and conservative because maybe a binary dilation could be applied to ensure no brain tissue is excluded).

By the same argument that *nifreeze should not deal with the generation of good/customized b=0s*, I don't think it should take responsibility for the mask either. If something is given from the exterior, the responsibility for storing it when/if necessary should rely on the exterior.
 
> * This block
>     https://github.com/nipreps/nifreeze/blob/e964b967ddcd5cc7d9e5d98adb8937185bd88827/src/nifreeze/data/dmri.py#L390-L394
>     should be put to a function so that it can be re-used and changed should another strategy be preferred. To be done in a separate PR.

oesteban:
My code suggestions are at jhlegarreta/nifreeze#1

One issue I realized of while reviewing is that, with `insert_b0=False` the b- vecs/vals would not be written out!

Hopefully my suggestions make sense :)

jhlegarreta:
@oesteban I am fine with this if you believe this is the way forward. I'd grateful if you merged this as soon as I make the tests pass: we are back to the Stanford dataset issues :man_facepalming: (x-ref issue #149) :
https://github.com/nipreps/nifreeze/actions/runs/15519578287/job/43690979304?pr=82#step:11:222
```
  E   dipy.data.fetcher.FetcherError: The downloaded file, /home/runner/.dipy/stanford_hardi/HARDI150.nii.gz, does not have the expected md5
  E      checksum of "0b18513b46132b4d1051ed3364f2acbc". Instead, the md5 checksum was: "24f8bbdb75cbd20fd79b099a10f58567". This
  E      could mean that something is wrong with the file or that the upstream file has been
  E      updated. You can try downloading the file again or updating to the newest version of
  E      dipy.
```

PR #83, opened 2025-02-26
ENH: Prefer using NumPy's random number generator

jhlegarreta:
- ENH: Prefer using NumPy's random number generator
- STYLE: Use the `GradientTable` class for type hinting
- STYLE: Remove `ClassVar[dict]` type hinting from GPR param constraints
- ENH: Ignore `scikit-learn` `Interval` and `StrOptions` type checking
- MAINT: Require `nitransforms` 22.x

PR #84, opened 2025-03-19
DOC: Update main flowchart and make modality-agnostic

oesteban:
nifreeze-flowchart

### DIFF
--- a/docs/developers.rst
+++ b/docs/developers.rst
@@ -28,6 +28,7 @@ Information on specific functions, classes, and methods.
 .. toctree::
    :glob:
 
+   api/nifreeze.analysis
    api/nifreeze.cli
    api/nifreeze.data
    api/nifreeze.data.dmri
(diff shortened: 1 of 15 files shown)


jhlegarreta (commented):
Thanks for doing this @oesteban :100:.

Some comments:
- "Estimate of the misalignment (...)" : "Estimate misalignment (...)".
- "Model" in "Drop-in Model" and "Mapping"  in "Spatial Mapping" have their first letters capitalized: none of the rest of titles have them. I'd use lowercase for these as well.
- The "modality-specific structure box" is unclear to me: the distinction between "Baseline artifact" and "Artifact estimate" is unclear: the name suggest they are an artifact, but the text is about transforms. Can this be clarified/put some other way?

jhlegarreta:
@oesteban `type-check` errors are being addressed in PR #83.

jhlegarreta (approved):

jhlegarreta (commented):
While fixing the font stuff, maybe you can use `held-out` instead of `held out` (we've used the former in the RR)

jhlegarreta (approved):

PR #86, opened 2025-03-30
ENH: Add Bland-Altman plot function

jhlegarreta:
Add Bland-Altman plot function.

Includes the ability to highlight N lower and upper extreme values along the X axis.

### DIFF
--- a/src/nifreeze/viz/bland_altman.py
+++ b/src/nifreeze/viz/bland_altman.py
@@ -1,0 +1,136 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright The NiPreps Developers <nipreps@gmail.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+"""Bland-Altman plot."""
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from nifreeze.analysis.measure_agreement import (
+    compute_bland_altman_features,
+    compute_z_score,
+)
+
+
+def plot_bland_altman(
+    data1: np.ndarray,
+    data2: np.ndarray,
+    ci: float = 0.95,
+    figsize: tuple | None = (15, 10),
+) -> plt.Figure:
+    """Create a Bland-Altman plot.
+
+    Create a Bland-Altman plot [Bland86]_ and highlight ``size`` lower and upper
+    extremes along the X coordinates.
+
+    Parameters
+    ----------
+    data1 : :obj:`numpy.ndarray`
+        Data values.
+    data2 : :obj:`numpy.ndarray`
+        Data values.
+    ci : :obj:`float`, optional
+        Confidence interval value. Must be in the [0, 1] range.
+    figsize : :obj:`tuple`, optional
+        Figure size.
+
+    Returns
+    -------
+    fig : :obj:`matplotlib.pyplot.Figure`
+        Matplotlib figure.
+
+    References
+    ----------
+    .. [Bland86] J. Martin Bland and Douglas G. Altman, Statistical methods for
+       assessing agreement between two methods of clinical measurement, The
+       Lancet 327(8476) (1986) 307-310
+
+    """
+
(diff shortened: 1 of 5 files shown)


jhlegarreta:
Test result:

src/nifreeze/viz/bland_altman.py
```diff
@@ -0,0 +1,284 @@
...
+#
+"""Bland-Altman plot."""
+
+import matplotlib.pyplot as plt
+import numpy as np
+import scipy.stats as stats
+
+
+def _compute_z_score(ci):
+    """Compute the z-score for the given confidence interval value.
```
arokem, line 31:
```suggestion
    """Compute the critical z-score for being outside a confidence interval.
```

src/nifreeze/viz/bland_altman.py
```diff
@@ -0,0 +1,284 @@
...
+
+    Parameters
+    ----------
+    ci : :obj:`float`
+        Confidence interval value.
```
arokem, line 36:
```suggestion
        Confidence interval size (number between 0 and 1).
```

src/nifreeze/viz/bland_altman.py
```diff
@@ -0,0 +1,284 @@
...
+    """Compute quantities of interest for the Bland-Altman plot.
+
+    Parameters
+    ----------
+    data1 : :obj:`numpy.ndarray`
+        Data values.
+    data2 : :obj:`numpy.ndarray`
+        Data values.
+    ci : :obj:`float`
+        Confidence interval value.
```
arokem, line 60:
```suggestion
        Confidence interval size (between 0 and 1).
```

src/nifreeze/viz/bland_altman.py
```diff
@@ -0,0 +1,284 @@
...
+
+    Returns
+    -------
+    diff : :obj:`numpy.ndarray`
+        Differences.
+    mean_diff : :obj:`numpy.ndarray`
+        Mean differences.
+    mean : :obj:`numpy.ndarray`
+        Mean values.
```
arokem, line 69:
```suggestion
        Mean values (across both data arrays).
```

src/nifreeze/viz/bland_altman.py
arokem: Or is it?

src/nifreeze/viz/bland_altman.py
```diff
@@ -0,0 +1,284 @@
...
+
+    n = len(diff)
+
+    # Compute t-value for the confidence intervals (CIs)
+    t_val = stats.t.ppf((1 + ci) / 2, n - 1)
+
+    # Compute confidence intervals
+
+    # Confidence interval for the mean difference
+    std_err_mean = std_diff / np.sqrt(n)
```
arokem, line 104: Shouldn't ddof for SEM be n-1?

arokem: Could also rely on scipy stats to compute this: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.sem.html

src/nifreeze/viz/bland_altman.py
```diff
@@ -0,0 +1,284 @@
...
+
+    # Compute confidence interval limits of agreement (LoA)
+    z_score = _compute_z_score(ci)
+    loa_lower = mean_diff - z_score * std_diff
+    loa_upper = mean_diff + z_score * std_diff
+
+    n = len(diff)
+
+    # Compute t-value for the confidence intervals (CIs)
+    t_val = stats.t.ppf((1 + ci) / 2, n - 1)
```
arokem, line 99: Might be clearer as: 

```suggestion
    t_val = stats.t.ppf(0.5 + ci / 2, n - 1)
```

arokem (commented):
Very nice! Some small comments

src/nifreeze/viz/bland_altman.py
jhlegarreta: It is.

src/nifreeze/viz/bland_altman.py
jhlegarreta: `std_err_mean = stats.sem(diff, ddof=0)` would give the same result as `std_err_mean = std_diff / np.sqrt(n)`; `std_err_mean = stats.sem(diff, ddof=1)` (default) would give the same result as `std_err_mean = std_diff / np.sqrt(n-1)`.

TBH, I do not know what is the correct way in this case, as it is the case here ase well https://github.com/nipreps/nifreeze/pull/86/files#diff-f5c68a621d0e143caebb0191455696dc29900bc9316ba28f843951ec0432bdb8R89

jhlegarreta:
I realize that in commit 8837386 (and previous ones) the extreme point identification did not honor how we intend to identify these, and I missed the part where we identify the reliable data (and color them differently in the BA plot). Also, this should be outside the visualization function, so that the identified data can be used and characterized separately.

I am marking this as a WIP until I figure out the best way to do this.

jhlegarreta:
@oesteban IMHO this is ready to be merged:
- Re https://github.com/nipreps/nifreeze/pull/86#discussion_r2020804405: this can be considered in a separate PR, but I think that the way it is done now is good/consistent.
- Re https://github.com/nipreps/nifreeze/pull/86#issuecomment-2767684823: the salient point identification is being addressed n PR #87.

oesteban (approved):
LGTM!!

PR #87, opened 2025-04-02
ENH: Identify Bland-Altman salient points

jhlegarreta:
Identify Bland-Altman salient points.

### DIFF
--- a/src/nifreeze/analysis/measure_agreement.py
+++ b/src/nifreeze/analysis/measure_agreement.py
@@ -22,8 +22,19 @@
 #
 """Measure agreement computation."""
 
+import enum
+
 import numpy as np
 import scipy.stats as stats
+
+
+class BASalientEntity(enum.Enum):
+    RELIABILITY_INDICES = "reliability_indices"
+    RELIABILITY_MASK = "reliability_mask"
+    LEFT_INDICES = "left_indices"
+    LEFT_MASK = "left_mask"
+    RIGHT_INDICES = "right_indices"
+    RIGHT_MASK = "right_mask"
 
 
 def _check_ci(ci: float) -> None:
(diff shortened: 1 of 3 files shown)


src/nifreeze/analysis/measure_agreement.py
```diff
@@ -0,0 +1,226 @@
...
+    ) = compute_bland_altman_features(data1, data2, ci)
+
+    # Filter data within the CI
+    reliable_mask = get_reliabilty_mask(diff, loa_lower, loa_upper)
+    reliability_idx = np.where(reliable_mask)[0]
+    reliable_mean = mean[reliable_mask]
+
+    # Compute the ratio between max and median of mean values
+    median_mean = np.median(reliable_mean)
+    mean_ratios = reliable_mean / median_mean
```
jhlegarreta, line 185: This needs to be reviewed. I do not think my understanding of the points that we seek here is correct.

jhlegarreta:
Result:

jhlegarreta:
@oesteban Re https://github.com/nipreps/nifreeze/pull/87#issuecomment-2842440544:

> - We want to create a set of BOLD runs that are most likely to be low-motion. This would be the left side of the BA plot (meaning the mean FD calculated with MCFLIRT and 3dVolReg is low, which most of the times will happen because both are low).

:+1: And this regardless of the difference or do we also enforce the identified points to be those in which *both* tools have agreed (i.e. around the mean difference)? If we want points to be around the mean difference, what criteria do you propose to use? e.g. 20% around the mean?

> - We want to identify datasets where motion is high. In this case this is the right-side of the BA plot, but probably we want something more reduced around the zero-difference line. Meaning, we want to identify examples where 3dVolreg and MCFLIRT agreed on the dataset having large head motion.

Linked to the above, this would be those within e.g. 20% of the mean difference value?

The above subsets are clear, but linked to https://github.com/nipreps/nifreeze/pull/87/files#r2023940688, the RR drafts says something else, so things are confusing. Please have a look at that; we can discuss it live.

jhlegarreta:
@oesteban please check if this is in agreement with the latest thoughts about the identification. For now, the rightmost number of points are less than `top_n` because they I am a taking the subset that is around the zero difference, but the concept is there.

bland-altman_salient_data

oesteban:
> Once I have identified the N leftmost points, I leave them aside, then sort the remaining ones starting from the right, then take 75% of these (to discard the central ones), then take N around the zero mean difference value.

I think this is great! Thanks!

Please mark it ready for review it does feel ready.

jhlegarreta:
> Please mark it ready for review it does feel ready.

Done.

oesteban (approved):

PR #89, opened 2025-04-23
ENH: Add a script to query fMRI compliant data from OpenNeuro

jhlegarreta:
Add a script to query fMRI compliant data from OpenNeuro.

### DIFF
--- a/scripts/query_fmri_data_openneuro.py
+++ b/scripts/query_fmri_data_openneuro.py
@@ -1,0 +1,406 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright The NiPreps Developers <nipreps@gmail.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+"""Select fMRI data from OpenNeuro based on a seed value and the following
+constraints:
+- A total of 4000 runs across datasets
+- No single dataset contributes to more than 5% of the total number of runs (200 runs)
+- BOLD run has between 300 and 1200 timepoints
+"""
+
+import argparse
+import logging
+
+import numpy as np
+import requests
+import re
+from datalad import api as dl
+from pathlib import Path
+
+import nibabel as nib
+import pandas as pd
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+
+OPENNEURO_DATASETS_URL = "https://github.com/OpenNeuroDataset"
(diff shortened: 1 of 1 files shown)


jhlegarreta:
In 4531431 I added the script to query the number of timepoints per BOLD run. Only missing the script to identify the runs that comply with the required criteria.

Before going forward, I'd like to check whether we are happy with the proposed approach

jhlegarreta:
Re https://github.com/nipreps/nifreeze/pull/89#issuecomment-2989171384

> Next is to develop/refactor the necessary parts for extracting the relevant features and selecting those datasets/files complying with the criteria into different scripts.

Done in subsequent commits.

I think I am done with this. Thoughts:
 - __Some datasets are no longer BIDS-compliant__ (visible as errors/warnings when browsing OpenNeuro) or seem to have some issues. Do we want to identify those? How can we identify these programmatically?
 - I am __only keeping__ those datasets that have __`human`__ in their `species` field. __Most__ datasets __have the `species` field empty__ in the response received from GraphQl. Not sure if this field is the same as in the JSON sidecar. Should we fetch those where the tag is empty? Should we double-check with the JSON sidecar? Looking for e.g. "mouse", "mice", "rat", macaque", "non-human primate" in the description sounds brittle. If we decide to include any of these a visual QA (if an automatic one does not exist) should be done.
- As opposed to the initial approach in https://github.com/nipreps/nifreeze/pull/89/commits/ea8ee4c1cf7a8719207aa1c598a52b21dd7bfae4#diff-c8a4bb6425ad95a7dd63f31feb2f1ef0a676b7fd2659e6cb16d4d91891cd7e40R184 and https://github.com/nipreps/nifreeze/pull/89/commits/ea8ee4c1cf7a8719207aa1c598a52b21dd7bfae4#diff-c8a4bb6425ad95a7dd63f31feb2f1ef0a676b7fd2659e6cb16d4d91891cd7e40R245, I am exclusively __relying on the file list and shape of the NIfTI header__ instead of the JSON sidecar to get the __number of runs/timepoints__.
- I am __not checking__ whether there is some meaningful __fMRI task__, e.g. https://github.com/nipreps/nifreeze/pull/89/commits/ea8ee4c1cf7a8719207aa1c598a52b21dd7bfae4#diff-c8a4bb6425ad95a7dd63f31feb2f1ef0a676b7fd2659e6cb16d4d91891cd7e40R270

Additionally, as I've split all this into 4 different scripts, I think it has become apparent that most of the __scripts in__ the __`scripts`__ folder do not belong here, and should be put into a __separate repository__ where all scripts to produce context results should dwell, keeping here only those scripts that are necessary for the package to run, if any.

jhlegarreta:
Re: the commit message 4f9b1e9, re-run earlier commits again; 9f66a50441d5aff82f011ca3a6119876ca89107f still incurs a deadlock; 2567252072551aafe3a8d1b9b1da5f812ecd2901 completes  `query_mri_dataset_files_openneuro.py` to completion (without handling errors), so I will try to bring changes one by one to `query_mri_dataset_files_openneuro.py` and see where the deadlock starts to happen. Thanks for the patience.

jhlegarreta:
> Considering that all these scripts are heavily related, and the shared static variables, would it make sense to aggregate all of them into a single script, and select the behavior with click commands?

I like the idea, but I'd propose to do it in a separate PR in order to have a working version at least. I know merging this clutters the repository, and in other circumstances, I'd be for adopting the approach here, but in this case I believe that I've spent too much time on this PR, so I'd rather have something merged and work from there. Also, as said, I think these scripts, together with the rest of the scripts that are not essential for the package, should dwell in a separate repository. So I'd vote for merging this as is to have a record of something that works, then transferring to another repository and converting to a `click`-based script in that repository.

jhlegarreta:
> I missed having examples in the initial docstring of every file.

Examples added.

jhlegarreta:
So, feel free to open a new repository to specifically host only these scripts. I will push the scripts to there when that is done, then work on the `click` re-implementation, then (mid- to long-term) to generalizing to any modality.

PR #91, opened 2025-05-01
STYLE: Increase consistency in data path across notebooks

jhlegarreta:
Increase consistency in data path across notebooks: check if the `TEST_DATA_HOME` environment variable if set and use it if such is the case; otherwise, use `Path.home() / "nifreeze-tests"`.

This matches the convention used for the `dmri_covariance.ipynb` notebook.


jhlegarreta:
I am running the fMRI notebook to check that it is still working as agreed yesterday.

PR #94, opened 2025-05-01
FIX: Streamline the most basic usecase (bold data)

oesteban:
Ensures the CLI can load a NIfTI/h5 file and run the average model.


jhlegarreta (commented):
Can the PR message be put into the commit message, please?

jhlegarreta (approved):

oesteban:
Yes, waiting for tests to pass

oesteban:
I'm removing `test_main.py` because at this point the cli is not ready to do the right thing depending on the data incoming. I'll work on this tomorrow.

PR #98, opened 2025-05-01
ENH: Make data loader smarter

oesteban:


src/nifreeze/data/base.py
```diff
@@ -27,24 +27,21 @@
 from collections import namedtuple
 from pathlib import Path
 from tempfile import mkdtemp
-from typing import Any, Generic
+from typing import Any, Generic, Self
```
jhlegarreta, line 30:
```suggestion
from typing_extensions import Self
from typing import Any, Generic
```

jhlegarreta:
`codecov` failures are unrelated. Merging.

PR #99, opened 2025-05-01
ENH: Add framewise displacement peak detection function

jhlegarreta:
Add framewise displacement peak detection function.

### DIFF
--- a/src/nifreeze/analysis/filtering.py
+++ b/src/nifreeze/analysis/filtering.py
@@ -1,0 +1,52 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright The NiPreps Developers <nipreps@gmail.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+"""Analysis data filtering."""
+
+import numpy as np
+
+
+def normalize(x: np.ndarray):
+    r"""Normalize data using the z-score.
+
+    The z-score normalization is computed as:
+
+    .. math::
+
+        z_i = \frac{x_i - \mu}{\sigma}
+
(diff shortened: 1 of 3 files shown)


test/test_fd_analysis.py
oesteban, line 1: should we remove this new empty file?

jhlegarreta: Yes, sorry, my bad. Removed with the last push force.

jhlegarreta:
Still not happy with the module naming; suggestions are welcome.

src/nifreeze/analysis/motion_analysis.py
```diff
@@ -0,0 +1,57 @@
...
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+"""Motion analysis."""
+
+import numpy as np
+
+from nifreeze.analysis.filtering import normalize
```
oesteban, line 27:
```suggestion
from scipy.stats import zscore
```

src/nifreeze/analysis/motion_analysis.py
```diff
@@ -0,0 +1,57 @@
...
+    Returns
+    -------
+    indices : :obj:`~numpy.ndarray`
+        Indices of identified motion spikes.
+    mask : :obj:`~numpy.ndarray`
+        Mask of identified motion spikes.
+    """
+
+    # Normalize (z-score)
+    fd_norm = normalize(fd)
```
oesteban, line 52:
```suggestion
    fd_norm = zscore(fd)
```

test/test_analysis.py
```diff
@@ -25,12 +25,14 @@
 import numpy as np
 import pytest
 
+from nifreeze.analysis.filtering import normalize
```
oesteban, line 28: No need to test this if using scipy
```suggestion
```

test/test_analysis.py
```diff
@@ -141,3 +143,34 @@ def test_identify_bland_altman_salient_data():
...
+    rng = request.node.rng
+
+    n_samples = 450
+
+    x = rng.normal(0, 5, n_samples)
+    expected_z_score = (x - np.mean(x)) / np.std(x)
+
+    obtained_z_score = normalize(x)
+
+    assert np.allclose(obtained_z_score, expected_z_score)
```
oesteban, line 158: Not necessary if using scipy:

```suggestion
```

src/nifreeze/analysis/filtering.py
oesteban, line 1: Let's use scipy for z-scoring instead of reimplementing it :)

src/nifreeze/analysis/motion_analysis.py
oesteban, line 1: since the module is `nifreeze.analysis` I think this module could better be `nifreeze.analysis.motion` and it's much easier than `motion_analysis`

jhlegarreta: :+1: Moved contents to the existing `motion.py` module in 5ee9efe1736deead2ad99a250eccd08c9bb0bc02. Force pushed..

jhlegarreta:
Codecov failures are unrelated. Merging.

oesteban:
The `filtering` module can be removed right?

PR #100, opened 2025-05-02
FIX: Actually run notebooks on CI

jhlegarreta:
Fix running notebooks:
- Notebooks were being supposedly run within the `test.yml` workflow. However, the fMRI BOLD notebook requires some data to be downloaded from OpenNeuro, and the data was not being downloaded.
- Run the data downloading script within the `notebooks` testing environment in `tox.ini`
- Run notebooks once every week through a cron job in a dedicated workflow file.
- Modify the Python versions being used in the CI `test.yml` file to avoid running the notebooks.

### DIFF
--- a/.github/workflows/notebooks.yml
+++ b/.github/workflows/notebooks.yml
@@ -1,0 +1,30 @@
+name: Run notebooks
+
+on:
+  schedule:
+    - cron: '0 0 * * 1'  # Every Monday at 00:00 UTC
+
+jobs:
+  run-notebooks:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        python-version: ['3.12']
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install tox
+        run: |
+          python -m pip install --upgrade pip
+          pip install tox
+
+      - name: Run notebooks with tox
+        run: tox -e notebooks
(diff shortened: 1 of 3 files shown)


oesteban (approved):
Thanks for this!

.github/workflows/notebooks.yml
```diff
@@ -0,0 +1,31 @@
...
+  #  - cron: '0 0 * * 1'  # Every Monday at 00:00 UTC
+  workflow_dispatch:     # Optional: allow manual triggering
+
+jobs:
+  run-notebooks:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        python-version: ['3.12-notebooks']
```
effigies, line 14:
```suggestion
        python-version: ['3.12']
```

.github/workflows/notebooks.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version.replace('-notebooks', '') }}
```
effigies, line 23:
```suggestion
          python-version: ${{ matrix.python-version }}
```

tox.ini
```diff
@@ -11,8 +11,9 @@ skip_missing_interpreters = true
 python =
   3.10: py310
   3.11: py311
-  3.12: py312, notebooks
+  3.12: py312
   3.13: py313
+  3.12-notebooks: py312, notebooks
```
effigies, line 16:
```suggestion
```

effigies (commented):
I think these three changes should work.

jhlegarreta:
@oesteban I think this is ready to go; I will keep an eye to see if the scheduled task does what is expected. Unless you want to try before merging (I did not succeed in making the workflow run even if requiring it on `push`, `pull_request` to `main`).

Thanks Chris.

PR #106, opened 2025-05-03
TST: Add back CLI main entry point testing

### LINKED ISSUE #101: Add back testing the CLI main entry point
The CLI main entry point testing was removed in PRs #94 and #96 as it was not covering all required use cases. See
 https://github.com/nipreps/nifreeze/pull/94#issuecomment-2845767319.

Testing the CLI main entry point is essential towards ensuring that the delivered package provides a simple-to-use interface.

After the refactoring done in PR #98, we should be able to add back the test covering the expected use cases.

jhlegarreta:
Add back CLI main entry point testing.

Fixes #101.


jhlegarreta:
@oesteban Not sure whether you want to add more testing to this now or whether that can be left for future work.

jhlegarreta:
I see. My bad.

And we do not want to keep the following?
https://github.com/nipreps/nifreeze/pull/106/files#diff-959a1effd1cb0e25016584deef9be1994af65814f318346bf3a63801b5db1636R32-R36

Re https://github.com/nipreps/eddymotion/pull/184.

> Rather than testing the CLI, what's missing is separating the parsing from the "main" function. That way, we will be able to offer that is usable to other libraries, as opposed to the current implementation that includes the parsing in the begining of the function.

So you mean changing
https://github.com/nipreps/nifreeze/blob/e488df25bac769d73f2e4d4772bf370aefc6f77e/src/nifreeze/cli/run.py#L32-L42

to?

```
def main(args) -> None: 
    (...)


if __name__ == "__main__":
    parser = build_parser()
    args = parser.parse_args(argv) 
    main(args)
```

src/nifreeze/cli/run.py
```diff
@@ -67,16 +48,11 @@ def main(argv=None) -> None:
...
     prev_model: Estimator | None = None
     for _model in args.models:
-        single_fit = _model.lower().startswith("single")
+        single_fit = estimator_kwargs[_model]["single_fit"]
         estimator: Estimator = Estimator(
-            _model.lower().replace("single", ""),
+            _model,
             prev=prev_model,
             single_fit=single_fit,
             model_kwargs=model_kwargs,
```
jhlegarreta, line 58: The estimator accepts other keyword arguments:
https://github.com/nipreps/nifreeze/blob/94f6d2708d15fbe85fb850094e956b63d9c047b9/src/nifreeze/estimator.py#L84

To be fixed in a separate PR.

src/nifreeze/cli/run.py
```diff
@@ -92,12 +68,14 @@ def main(argv=None) -> None:
...
-    output_filename: str = Path(args.input_file).name
+    output_filename = Path(Path(args.input_file).name).stem + ".nii.gz"
     output_path: Path = Path(args.output_dir) / output_filename
 
     # Save the DWI dataset to the output path
     if args.write_hdf5:
-        dataset.to_filename(output_path)
+        output_h5_filename = Path(Path(args.input_file).name).stem + ".h5"
+        output_h5_path: Path = Path(args.output_dir) / output_h5_filename
+        dataset.to_filename(output_h5_path)
```
jhlegarreta, line 78: Although having an HDF5 file as input seems the best way for testing,
https://github.com/nipreps/nifreeze/blob/94f6d2708d15fbe85fb850094e956b63d9c047b9/src/nifreeze/cli/parser.py#L68

In the mid-term we will want to accept NIfTI files. So the above will need to change.

To be addressed in a separate PR.

src/nifreeze/cli/run.py
jhlegarreta, line 58: The estimator accepts other keyword arguments:
https://github.com/nipreps/nifreeze/blob/94f6d2708d15fbe85fb850094e956b63d9c047b9/src/nifreeze/estimator.py#L84

To be fixed in a separate PR.

jhlegarreta:
The above comments (except for the [monkeypatching question](https://github.com/nipreps/nifreeze/pull/106/files#r2147086912)) are left for the records and for separate/future PRs. Would be grateful if this could be reviewed and potentially merged soon unless I got it wrong.

test/test_parser.py
```diff
@@ -88,40 +137,147 @@ def test_models_arg(tmp_path, argval, _models):
...
+
+    if ignore_b0:
+        assert model_kwargs["ignore_bzero"] == ignore_b0
+
+
+@pytest.mark.parametrize(
+    "models",
+    [
+        "dti",
+        "dki",
```
jhlegarreta, line 220: Did not include `averagedwi`:
https://github.com/nipreps/nifreeze/blob/7cfc148cdcf9ab24551054bdafdbefbf04539ec3/src/nifreeze/model/base.py#L61
because I am testing here the `_model_class` property:
https://github.com/nipreps/nifreeze/pull/106/files#diff-240d2e4cff5653f9b0b21aaf6d06a74165c1b72f21c0717735d5c26b41bd52c6R251

which `AverageDWIModel` does not have.

The `TrivialModel` and `ExpectationModel` (to be done in a separate function) were not added due to the same reason.

Maybe some other property could be tested and this function be refactored. To be done in a separate PR.

Slightly related, even if an `S0` model is mentioned in https://github.com/nipreps/nifreeze/blob/7cfc148cdcf9ab24551054bdafdbefbf04539ec3/src/nifreeze/model/base.py#L47, it is not implemented or its alias not included elsewhere.

src/nifreeze/cli/parser.py
```diff
@@ -164,3 +164,106 @@ def build_parser() -> ArgumentParser:
...
+    model_name : :obj:`str`
+        Model name.
+
+    Returns
+    -------
+    :obj:`str`
+        Normalized model name.
+    """
+
+    return model_name.lower().replace("single", "")
```
oesteban, line 207: IMHO, spawning one-liner string manipulations into functions is less transparent

test/conftest.py
```diff
@@ -285,7 +285,7 @@ def setup_random_dwi_data(request, setup_random_gtab_data):
     dwi_dataobj, affine = _generate_random_uniform_spatial_data(
         request, (*vol_size, volumes), 0.0, 1.0
     )
-    brainmask_dataobj = rng.random(vol_size, dtype="float32")
+    brainmask_dataobj = rng.choice([True, False], size=vol_size).astype(np.uint8)
```
oesteban, line 288: We should do this on a separate PR.

test/data/hcph_multishell.txt
```diff
@@ -0,0 +1,4 @@
...2000.000000 3000.000000 2000.000000 3000.000000 2000.000000 1000.000000 3000.000000 2000.000000 0.000000 3000.000000 1000.000000 3000.000000 3000.000000 2000.000000 2000.000000 700.000000 3000.000000 3000.000000 2000.000000 3000.000000 700.000000 3000.000000 2000.000000 2000.000000 3000.000000 2000.000000 2000.000000 3000.000000 2000.000000 3000.000000 3000.000000 2000.000000 3000.000000 2000.000000 3000.000000 1000.000000 1000.000000 3000.000000 1000.000000 2000.000000 700.000000 2000.000000 1000.000000 2000.000000 3000.000000 3000.000000 700.000000 3000.000000 2000.000000 3000.000000 1000.000000 3000.000000 2000.000000 1000.000000 2000.000000 0.000000 3000.000000 1000.000000 1000.000000 3000.000000 3000.000000 3000.000000 1000.000000 3000.000000 2000.000000 2000.000000 2000.000000 3000.000000 3000.000000 2000.000000 2000.000000 3000.000000 700.000000 3000.000000 3000.000000 2000.000000 1000.000000 2000.000000 3000.000000 3000.000000 3000.000000 2000.000000 3000.000000 3000.000000 3000.000000 1000.000000 3000.000000 2000.000000 1000.000000 2000.000000 2000.000000 2000.000000 3000.000000 3000.000000 3000.000000 1000.000000 3000.000000 3000.000000 3000.000000 3000.000000 3000.000000
```
oesteban, line 1: Do we need this file? If we're not testing other than the CLI implementation, an empty file or a file with random content should be good enough, right?

oesteban (commented):
A few nit picks

src/nifreeze/cli/parser.py
jhlegarreta: OK, now I see another reason why that went into a separate function:
https://github.com/nipreps/nifreeze/actions/runs/16146097132/job/45565241798?pr=106#step:11:671

Are we OK if I bring them back, and have this PR merged? I do not want to go over more failures...

test/conftest.py
jhlegarreta: Resolving. Let's have this fix integrated in the `main` branch.

jhlegarreta:
Ready to be merged. Improvements can come in separate PRs.

PR #113, opened 2025-05-04
ENH: Add NuFO computation script

jhlegarreta:
Add NuFO computation script.


jhlegarreta:
This will be useful when we will want to plot the GP estimation vs the DTI estimation on voxels having an arbitrary number of peaks.

jhlegarreta:
The brain mask is conservative/not dilated since we will want to make sure that we select voxels from the brain.

oesteban (approved):

PR #115, opened 2025-05-08
TST: Download dMRI covariance notebook data from OSF

jhlegarreta:
Download dMRI covariance notebook data from OSF instead of using G-Node GIN: the latter is orders of magnitude slower, which is inconvenient to test the notebook.

The data has been previously uploaded to OSF.

Install the `osfclient` in order to effectively download the data.


jhlegarreta:
Benchmark CI issue is unrelated.

Merging.

jhlegarreta:
PR to address benchmark CI issue: #116.

PR #117, opened 2025-05-08
TST: Do not remove fMRI notebook from list of discovered notebooks

jhlegarreta:
Do not remove fMRI notebook from list of discovered notebooks: this will allow running the notebook at the time scheduled by the `notebooks.yml` GHA workflow file.

PR #121, opened 2025-05-14
FIX: Fix FD plot function call in BOLD fMRI realignment notebook

jhlegarreta:
Fix FD plot function call in BOLD fMRI realignment notebook: provide the framewise displacement data as a `pandas` `DataFrame` and group the labels for the legend into a list.

Left behind in commit 2d95cbb.

Take advantage of the commit to transfer a lonely `import` statement to the immediately subsequent cell where the actual plot function is called for the sake of compactness and consistency.

### DIFF
--- a/docs/notebooks/bold_realignment.ipynb
+++ b/docs/notebooks/bold_realignment.ipynb
@@ -166,19 +166,13 @@
   },
   {
    "cell_type": "code",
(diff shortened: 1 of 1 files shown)


jhlegarreta:
Merging. We will only be able to test this on GH once we fix issues with PR #117.

PR #126, opened 2025-05-20
MAINT: Increase notebook timeout value

jhlegarreta:
Increase notebook timeout value: set it to 3600 s (60 minutes; the default value is 300 s).

Fixes:
```
__ /home/runner/work/nifreeze/nifreeze/docs/notebooks/bold_realignment.ipynb ___
A cell timed out while it was being executed, after 300 seconds.
The message was: Cell execution timed out.
```

raised for example at:
https://github.com/nipreps/nifreeze/actions/runs/15102456036/job/42445432169#step:15:43

### DIFF
--- a/tools/run_notebooks.py
+++ b/tools/run_notebooks.py
@@ -28,4 +28,4 @@ import sys
 # Find and run the notebooks
 notebooks = glob.glob("docs/notebooks/*.ipynb")
 
-sys.exit(subprocess.call(["pytest", "--nbmake"] + notebooks))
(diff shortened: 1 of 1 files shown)


jhlegarreta:
This is now doing its job: brainmasks are being computed and the failure is now about the issues with the missing cells/statements in the notebook:
https://github.com/jhlegarreta/nifreeze/actions/runs/15127176019/job/42521243652

Merging.

jhlegarreta:
Coming up with a more generalizable approach to set the timeout value is left for an eventual future PR.

PR #129, opened 2025-05-20
DOC: Miscellaneous improvements to documentation

jhlegarreta:
- DOC: Do not use a dash in the general case of the "head motion" term
- DOC: Honor better purpose in estimator module and class documentation

### DIFF
--- a/src/nifreeze/estimator.py
+++ b/src/nifreeze/estimator.py
@@ -67,7 +67,7 @@ class Filter:
 
 
 class Estimator:
-    """Estimates rigid-body head-motion and distortions derived from eddy-currents."""
+    """Estimates rigid-body head motion and distortions derived from eddy-currents."""
 
     __slots__ = ("_model", "_strategy", "_prev", "_model_kwargs", "_align_kwargs")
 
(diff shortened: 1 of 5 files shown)


src/nifreeze/estimator.py
```diff
@@ -20,7 +20,7 @@
 #
 #     https://www.nipreps.org/community/licensing/
 #
-"""A model-based algorithm for the realignment of dMRI data."""
+"""A model-based algorithm for head motion and eddy current distortion estimation and realignment for 4D neuroimaging data."""
```
oesteban, line 23: Description should be a one-liner, what about this?

```suggestion
"""Orchestrates model and registration in volume-to-volume artifact estimation."""
```

src/nifreeze/estimator.py
```diff
@@ -67,7 +67,7 @@ def run(self, dataset: DatasetT, **kwargs) -> DatasetT:
 
 
 class Estimator:
-    """Estimates rigid-body head-motion and distortions derived from eddy-currents."""
+    """Rigid-body head motion and eddy current distortion estimator for 4D neuroimaging data using a model-based approach."""
```
oesteban, line 70:
```suggestion
    """Orchestrates components for a single estimation step."""
```

PR #137, opened 2025-05-22
STY: Add license header to author update script

jhlegarreta:
Add license header to author update script.

Take advantage of the commit to remove the Python interpreter shebang line for the sake of consistency across the package and current conventions across the NiPreps projects.


jhlegarreta:
Not sure if
- Using the shebang line is still necessary/preferred.
- Not adding the license header was on purpose.

PR #138, opened 2025-05-22
MNT: Configure `coverage` for parallel execution

jhlegarreta:
Configure `coverage` for parallel execution: set `coverage`'s `parallel` flag to `true` so that when reporting coverage tests running in parallel (achieved using `pytest-xdist`, and enabled for `coverage` using the `multiprocessing` concurrency) do not fall into race conditions and write to different `.coverage.` files. These files need to be assembled later to provide aggregated statistics.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -229,6 +229,7 @@ filterwarnings = [
 
 [tool.coverage.run]
 branch = true
+parallel = true
 concurrency = ['multiprocessing']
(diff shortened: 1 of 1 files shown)


jhlegarreta:
We do not seem to have run into trouble (most of the times?), but just in case. Also, we do not call `combine`; so not sure how results are aggregated.

PR #139, opened 2025-05-23
FIX: Revise interface and implementation of PET model

oesteban:
I simplified the model interface after the refactor in #62.

### DIFF
--- a/src/nifreeze/model/pet.py
+++ b/src/nifreeze/model/pet.py
@@ -22,10 +22,11 @@
 #
 """Models for nuclear imaging."""
 
+from os import cpu_count
+
 import numpy as np
 from joblib import Parallel, delayed
 
-from nifreeze.exceptions import ModelNotFittedError
 from nifreeze.model.base import BaseModel
 
 DEFAULT_TIMEFRAME_MIDPOINT_TOL = 1e-2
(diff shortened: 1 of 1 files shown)


src/nifreeze/model/pet.py
```diff
@@ -101,26 +96,22 @@ def fit(self, data, **kwargs):
...
-
         # Parallelize process with joblib
-        with Parallel(n_jobs=n_jobs) as executor:
+        with Parallel(n_jobs=n_jobs or min(cpu_count() or 1, 8)) as executor:
             results = executor(delayed(cg)(ATdotA, AT @ v) for v in data)
 
         self._coeff = np.array([r[0] for r in results])
 
-    def predict(self, index=None, **kwargs):
+    def fit_predict(self, index: int | None = None, **kwargs):
```
oesteban, line 105: Preparing for #120 (this model is thought to be fit only once as it is designed)

src/nifreeze/model/pet.py
```diff
@@ -130,9 +121,5 @@ def predict(self, index=None, **kwargs):
         # self._coeff is V (num. voxels) x K - 4
         predicted = np.squeeze(A @ self._coeff.T)
 
-        if self._mask is None:
-            return predicted.reshape(self._datashape)
-
-        retval = np.zeros(self._datashape, dtype="float32")
-        retval[self._mask] = predicted
-        return retval
```
oesteban, line 138: The mask was not properly applied because the fit is done in the full dataset, it doesn't make sense to only apply the mask at the output. I'll leave fitting only within the mask for a future PR.

oesteban:
I'm going to move ahead and merge this as it is important to push it asap. I'm happy to help resolve potential conflicts with #88.

PR #143, opened 2025-05-24
ENH: Add GQI to available models

oesteban:
Enables Generalized Q-Imaging to the portfolio of DIPY supported models.

I'm hitting the issue that this model does not have a `predict()` implementation (cc/ @yasseraleman).

```
NotImplementedError: This model does not have prediction implemented yet
```

@arokem, would it be very hard to write?

### DIFF
--- a/src/nifreeze/model/base.py
+++ b/src/nifreeze/model/base.py
@@ -66,7 +66,7 @@ class ModelFactory:
 
             return AverageDWIModel(kwargs.pop("dataset"), **kwargs)
 
-        if model.lower() in ("dti", "dki", "pet"):
+        if model.lower() in ("gqi", "dti", "dki", "pet"):
             from importlib import import_module
 
             dmrimod = import_module("nifreeze.model.dmri")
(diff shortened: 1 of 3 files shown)


arokem:
Yes. IIUC, GQI is not really a model in the sense that parameters are fit
to data according to some equation. Rather, much like DSI from which it was
derived, it’s a way to compute an ODF from diffusion data. So in some sense
the model “prediction” would identical to the data. Is there a particular
reason to implement it for nifreeze?

arokem:
Gotcha, yes, that does seem like an interesting option. But doesn't that push the problem to fitting an adequate kernel? In particular, you'd have to define how the kernel behaves at higher b-values (presumably something non-Gaussian). So, you'd have to fit a kernel to data (or maybe I am missing something? @yasseraleman: how do you do this?) The conceptual issues aside, I think this would not be too hard to implement, maybe first as a sub-class here?

jhlegarreta (commented):
@oesteban Just remove the keyword decorators: those were introduced in DIPY when the tool moved to a keyword-only argument specification. Since in this case this is the first commit introducing this code, we do not need those.

Issue #149, opened 2025-05-31
Replace the Stanford HARDI test dataset


jhlegarreta:
Downloading the DIPY Stanford HARDI dataset, creating a local directory to host it, or its MD5 checksum are failing on and off.

Every time it fails, the CI workflows have to be re-triggered manually. It has been happening a few times per push lately.

To save us work, we should use another dataset for testing purposes.

jhlegarreta:
> Why is it failing?

Have not found a pattern. Possibilities:
- Some race condition (?), e.g.
https://github.com/nipreps/nifreeze/actions/runs/15357323502/job/43218947747#step:11:254
https://github.com/nipreps/nifreeze/actions/runs/15301631097/job/43043463683#step:11:250
https://github.com/nipreps/nifreeze/actions/runs/15229643356/job/42835974833#step:11:136
```
  ___________________ ERROR collecting test/test_motion_viz.py ___________________
  /opt/hostedtoolcache/Python/3.10.17/x64/lib/python3.10/pathlib.py:1175: in mkdir
      self._accessor.mkdir(self, mode)
  E   FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/.dipy/.lock-stanford_hardi'
  
  During handling of the above exception, another exception occurred:
  test/test_motion_viz.py:51: in <module>
      lock_folder.mkdir(parents=True)
  /opt/hostedtoolcache/Python/3.10.17/x64/lib/python3.10/pathlib.py:1180: in mkdir
      self.mkdir(mode, parents=False, exist_ok=exist_ok)
  /opt/hostedtoolcache/Python/3.10.17/x64/lib/python3.10/pathlib.py:1175: in mkdir
      self._accessor.mkdir(self, mode)
  E   FileExistsError: [Errno 17] File exists: '/home/runner/.dipy/.lock-stanford_hardi'
  --- generated xml file: /home/runner/work/nifreeze/nifreeze/test-results.xml ---
(...)
  =========================== short test summary info ============================
  ERROR test/test_motion_viz.py - FileExistsError: [Errno 17] File exists: '/home/runner/.dipy/.lock-stanford_hardi'
```

- `MD5` check failures due to something else:
https://github.com/nipreps/nifreeze/actions/runs/15357403507/job/43219172502#step:11:205
```
  ___________________ ERROR collecting test/test_motion_viz.py ___________________
  test/test_motion_viz.py:52: in <module>
      fetch_stanford_hardi()
[...]

PR #166, opened 2025-07-09
TST: Add Python 3.13 to the testing suite

jhlegarreta:
Add Python 3.13 to the testing suite and the versions for which the packaging is tested.

Python 3.13 was released in October 2024, and Python 3.14 will be released in October 2025, so it seems reasonable to start testing on Python 3.13.

Release notes:
https://www.python.org/downloads/release/python-3130/

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -87,7 +87,7 @@ jobs:
     needs: ['cache-test-data']
     strategy:
       matrix:
-        python-version: ['3.10', '3.11', '3.12']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
     steps:
       - uses: actions/checkout@v4
         with:


.github/workflows/pythonpackage.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
```
oesteban, line 19: Let's drop 3.10:

```suggestion
        python-version: ["3.11", "3.12", "3.13"]
```

oesteban (commented):
Considering how long it takes to get stuff green, I'd stick with just one Python release for continuous integration and test all versions only before releases.

jhlegarreta:
> Considering how long it takes to get stuff green

IMO, it is not all that bad for now: unit tests are around 12 minutes. The thing that mostly fails and which does not depend on what we've coded in a commit is downloading one of the test datasets (x-ref issue #149).

> I'd stick with just one Python release for continuous integration and test all versions only before releases.

I'm not sure what GitHub would do once we tag/release a version and it fails (besides not publishing it or uploading to PyPI?): i.e. how it would handle the new commits required to fix the issues. I assume we would need to manually overwrite the tag, etc. Does not seem the easiest path.

Something that would inform us more regularly would be better, e.g. weekly builds on the entire set of versions. Doing that should not be difficult.

Relevant documentation for events that trigger workflows: https://docs.github.com/en/actions/reference/events-that-trigger-workflows

oesteban:
> Something that would inform us more regularly would be better, e.g. weekly builds on the entire set of versions. Doing that should not be difficult.

Yes, this would be fantastic. With some fiddling, we could also set one python version as default (e.g., 3.12 today) and then parse the branch name and change it:

if you are working on a branch named `py3.13/test-whatever` (or `test-whatever-py3.13`, or whatever convention we agree upon), then the tests are run with the version in the branch name. This would help debug when an error outside the default python version occurs.

jhlegarreta:
PR #169.

oesteban:
Let's get this one merged as is and then we can optimize with #169 and further. Thanks

PR #168, opened 2025-07-11
STY: Fix miscellaneous style issues

jhlegarreta:
- STY: Remove redundant parentheses when returning a tuple
- STY: Avoid using flow-dependent variable outside context
- DOC: Remove personal reference in comment


jhlegarreta:
`codecov/patch` failure is unrelated. Merging.

PR #174, opened 2025-07-18
WIP: Enh/pet model fix

mnoergaard:
Following the suggestion from https://github.com/nipreps/nifreeze/pull/88#issuecomment-2982076459, this PR adds a PET model and estimator.


jhlegarreta (commented):
Had a quick look. I think I prefer to have this merged as is (as long as the `WIP` in the PR subject does not apply) and do separate PRs to improve it. This would avoid going through the rebasing process again.

One only thing I would do is to squash all commits into one before merging.

How does this sound @oesteban @mnoergaard?

jhlegarreta:
Squashed and merged. Thanks @mnoergaard for the hard work and @oesteban for the support. We can now improve or refactor the necessary parts with the confidence that all 3 modalities are being considered.

PR #36, opened 2024-12-22
ENH: Implement squared exponential covariance function

jhlegarreta:
Implement squared exponential covariance function.

PR #175, opened 2025-07-23
ENH: Implement ``MultiShellKernel``

### LINKED ISSUE #337: Extend GPR to the multi-shell case

Generalize the implementation to accept multiple b-values

Yes

_No response_

### LINKED ISSUE #456: GP hyperparameter optimization: fix log-space kernel gradients and make noise (σ²) tunable via WhiteKernel
## Summary

Two coupled problems in the GP model layer (`src/nifreeze/model/gpr.py`, `_dipy.py`):

### 1. Kernel gradients are computed in linear space (bug)

`SphericalKriging` and `ExponentialKriging` return `K_gradient` as the derivative w.r.t. the hyperparameters themselves (e.g. `dC/da`, `C`). But scikit-learn's `Kernel.theta` is **log-transformed**, and the L-BFGS optimizer expects gradients w.r.t. `log θ`. By the chain rule the correct value is `dK/d(log θ) = θ · dK/dθ` — i.e. each channel must be scaled by its hyperparameter value (for the scale `λ`, the channel is `λ·C = K`, not `C`).

A finite-difference check (perturb `theta`, clone the kernel, recompute `K`) does not match the analytical gradient on the current code, and matches to ~1e-11 after the fix.

**Impact:** `GaussianProcessModel.fit` runs with the default `fmin_l_bfgs_b` optimizer and `eval_gradient=True`, so the production single-shell GP has been optimizing `beta_a`/`beta_l` with incorrect gradients. This affects the LOVO prediction and hence motion estimates.

### 2. Measurement noise `σ²` cannot be tuned (enhancement)

`σ²` is currently passed as scikit-learn's fixed `alpha` nugget (`alpha=self.sigma_sq`). Because `alpha` is added outside the kernel, its kernel-gradient is zero and it is never optimized. Following :cite:`andersson_non-parametric_2015` (which estimates `σ²`), the noise should be folded into the covariance as a `WhiteKernel` so it is optimized alongside the other hyperparameters. `sigma_sq` then becomes the initial `noise_level`, and `alpha` is demoted to a small numerical jitter.

## Proposed scope

- Fix the log-space gradients in both Kriging kernels (+ update the gradient-asserting tests, add a finite-difference regression test).
- Single-shell: `GaussianProcessModel` builds `KernelType(...) + WhiteKernel(noise_level=sigma_sq)`, tunable by default.
- **Later (separate work):** per-shell noise `σ_i²` (Eq. 16) for the multi-shell case is *not* a stock `WhiteKernel` — it needs a small custom kernel that applies block-diagonal noise per shell from the b-value column.

## Notes

- The gradient fix is a prerequisite for the multi-shell kernel work (#337 / PR #175), whose orientation sub-kernel reuses `SphericalKriging`.
- Behavioral change: making noise tunable shifts single-shell GP predictions; existing tests assert shapes/finiteness and still pass.

oesteban:
## Summary

Adds a `MultiShellKernel` so a single Gaussian process can model **multi-shell** diffusion data, following Andersson & Sotiropoulos (2015). The kernel is the product of an angular (orientation) covariance and a radial (log-b) covariance (Eqs. 14–15), letting the GP borrow strength across shells instead of fitting each shell independently.

It plugs into the existing GP path: `GaussianProcessModel(kernel_model="multishell")` builds a spherical angular × RBF-on-log-b radial kernel. The design matrix now carries the b-value (`[gx, gy, gz, bval]`), which the multi-shell kernel selects internally; single-shell kernels drop that column and keep consuming orientations only.

## Depends on the GP kernel/noise fix (#457, fixes #456)

This is rebased on top of #457 and relies on it:

- `MultiShellKernel` returns **analytical gradients in scikit-learn's log-hyperparameter space**, the same convention #457 corrected for the single-shell kernels; the gradient-based optimizer would otherwise be inconsistent across the angular and radial factors.
- Measurement noise is tuned through the `WhiteKernel` term introduced in #457, and the multi-shell path fits through that same flow.

## Tests

- **Kernel correctness** — `test_multishellkernel_basic` (shape, symmetry, `diag` consistency, gradient shape), `test_multishellkernel_reduces_to_radial_when_orientations_identical` (reduces to the radial kernel, Eq. 15, when orientations match), and `test_multishellkernel_column_selection` (only the configured orientation/b-value columns affect the result).
- **Analytical gradients** — `test_kernel_gradient_matches_finite_differences` now also covers `MultiShellKernel`, checking analytical vs. finite-difference gradients in log space.
- **Guards** — `test_multishellkernel_nonpositive_bval` and `test_gp_model_multishell_rejects_b0` ensure b0 / non-positive b-values are rejected rather than producing `-inf`/`nan`.
- **Cloning** — `test_multishellkernel_clone_theta_roundtrip` verifies `clone_with_theta` preserves structure.
- **End-to-end** — `test_multishellkernel_gp`, `test_multishellkernel_gp_optimizes`, and `test_gp_model_multishell` fit and predict through `DiffusionGPR` / `GaussianProcessModel` over two shells.

### DIFF
--- a/src/nifreeze/model/_dipy.py
+++ b/src/nifreeze/model/_dipy.py
@@ -43,15 +43,8 @@ GP_JITTER = 1e-10
 """Small nugget kept on ``alpha`` for numerical stability."""
 
 
-def _cartesian_components(gtab: GradientTable | np.ndarray) -> np.ndarray:
-    """Return the design matrix ``[gx, gy, gz, bval]``.
-
-    The b-value is always appended as a trailing column (layout
-    ``[gx, gy, gz, bval]``) so multi-shell kernels can access it. Single-shell
-    kernels operate on orientations only and drop the b-value column downstream
-    (see :func:`_design_matrix`).
-    """
-
+def _btable_asarray(gtab: GradientTable | np.ndarray) -> np.ndarray:
+    """Return the design matrix ``[gx, gy, gz, bval]``."""
     if hasattr(gtab, "bvecs"):
         gtab = cast(GradientTable, gtab)
         return np.column_stack([gtab.bvecs, np.asarray(gtab.bvals)])
@@ -60,19 +53,6 @@ def _cartesian_components(gtab: GradientTable | np.ndarray) -> np.ndarray:
     if components.ndim == 1:
         components = components[np.newaxis, :]
     return components
-
-
-def _design_matrix(gtab: GradientTable | np.ndarray, kernel: Kernel | None) -> np.ndarray:
-    """Build the GP design matrix appropriate for ``kernel``.
-
-    The b-value is always available in the Cartesian components; a single-shell
-    kernel only consumes orientations, so its trailing b-value column is dropped
-    here (multi-shell kernels keep it, selecting columns via ``bval_index``).
-    """
-    X = _cartesian_components(gtab)
-    if not isinstance(kernel, MultiShellKernel) and X.shape[-1] > 3:
-        X = X[:, :3]
-    return X
 
 
 def gp_prediction(
@@ -106,7 +86,10 @@ def gp_prediction(
 
     """
 
-    X = _design_matrix(gtab, getattr(model, "kernel", None))
+    X = _btable_asarray(gtab)
+    # Single-shell kernels consume orientations only; drop the b-value column.
+    if not isinstance(getattr(model, "kernel", None), MultiShellKernel):
+        X = X[:, :3]
 
     # Check it's fitted as they do in sklearn internally
     # https://github.com/scikit-learn/scikit-learn/blob/972e17fe1aa12d481b120ad4a3dc076bae736931/\
@@ -225,11 +208,13 @@ class GaussianProcessModel(ReconstModel):
 
         """
 
-        # Extract b-vecs: scikit-learn wants (n_samples, n_features) where
-        # n_samples is the different diffusion-encoding gradient orientations.
-        # n_features is 3 (orientation only), or 4 ([gx, gy, gz, bval]) when a
-        # multi-shell kernel needs the b-value.
-        X = _design_matrix(gtab, self.kernel)
+        # scikit-learn wants (n_samples, n_features), where n_samples is the
+        # number of diffusion-encoding gradient orientations. n_features is 4
+        # ([gx, gy, gz, bval]) for the multi-shell kernel; single-shell kernels
+        # consume orientations only, so the b-value column is dropped.
+        X = _btable_asarray(gtab)
+        if not isinstance(self.kernel, MultiShellKernel):
+            X = X[:, :3]
 
         # Data must have shape (n_samples, n_targets) where n_samples is
         # the number of diffusion-encoding gradient orientations, and n_targets
--- a/src/nifreeze/model/_dipy.py
+++ b/src/nifreeze/model/_dipy.py
@@ -24,15 +24,18 @@
 
 from __future__ import annotations
 
+from typing import cast
+
 import numpy as np
 from dipy.core.gradients import GradientTable
 from dipy.reconst.base import ReconstModel
 from sklearn.gaussian_process import GaussianProcessRegressor
-from sklearn.gaussian_process.kernels import WhiteKernel
+from sklearn.gaussian_process.kernels import RBF, Kernel, KernelOperator, WhiteKernel
 
 from nifreeze.model.gpr import (
     DiffusionGPR,
     ExponentialKriging,
+    MultiShellKernel,
     SphericalKriging,
 )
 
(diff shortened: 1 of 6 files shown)


jhlegarreta:
@oesteban Should we favor this PR instead of #36?

jhlegarreta:
Notes:
- Benchmarking failure is unrelated to this patch set.
- As for the design, rather than a kernel operator, this looks a kernel on its own. 
- We probably need better tests or a notebook to convince ourselves that this is working (I confess that tests are too simplistic and driven by copilot).

src/nifreeze/model/_dipy.py
```diff
@@ -24,36 +24,63 @@
...
 
-def _cartesian_components(gtab: GradientTable | np.ndarray) -> np.ndarray:
-    """Return the gradient directions as Cartesian components."""
+def _cartesian_components(gtab: GradientTable | np.ndarray, keep_bval: bool = False) -> np.ndarray:
+    """Return the gradient directions as Cartesian components.
+
+    When ``keep_bval`` is :obj:`True`, the b-value is appended as a trailing
+    column (layout ``[gx, gy, gz, bval]``) so multi-shell kernels can access it;
+    otherwise only the 3 orientation components are returned.
+    """
```
oesteban: Wouldn't it make more sense to keep the b-vals always and let the single-shell GPR drop the b-vals column?

src/nifreeze/model/_dipy.py
```diff
@@ -125,8 +154,11 @@ def __init__(
         Parameters
         ----------
         kernel_model : :obj:`str`, optional
-            Angular covariance model, ``"spherical"`` (default) or
-            ``"exponential"``.
+            Angular covariance model. One of ``"spherical"`` (default),
+            ``"exponential"``, or ``"multishell"``. ``"multishell"`` builds a
```
oesteban, line 134: say that multishell is Spherical, and indicate whether Andersson allowed/investigated multishell with exponential

src/nifreeze/model/gpr.py
```diff
@@ -435,7 +455,9 @@ def __call__(
         if not eval_gradient:
             return self.beta_l * C_theta
 
-        # scikit-learn expects gradients w.r.t. the *log* of each hyperparameter.
+        # scikit-learn expects gradients w.r.t. the *log* of each hyperparameter
+        # (``Kernel.theta`` is log-transformed), hence the chain-rule factors of
+        # ``beta_a`` and ``beta_l``.
```
oesteban:
```suggestion
        # scikit-learn expects gradients w.r.t. the *log* of each hyperparameter
```

src/nifreeze/model/_dipy.py
```diff
@@ -41,19 +44,37 @@
...
 
 def _cartesian_components(gtab: GradientTable | np.ndarray) -> np.ndarray:
-    """Return the gradient directions as Cartesian components."""
+    """Return the design matrix ``[gx, gy, gz, bval]``.
+
+    The b-value is always appended as a trailing column (layout
+    ``[gx, gy, gz, bval]``) so multi-shell kernels can access it. Single-shell
+    kernels operate on orientations only and drop the b-value column downstream
+    (see :func:`_design_matrix`).
+    """
```
oesteban:
```suggestion
    """Return the design matrix ``[gx, gy, gz, bval]``."""
```

src/nifreeze/model/_dipy.py
```diff
@@ -41,19 +44,37 @@
 
 
 def _cartesian_components(gtab: GradientTable | np.ndarray) -> np.ndarray:
```
oesteban: name not representative anymore?

```suggestion
def _btable_asarray(gtab: GradientTable | np.ndarray) -> np.ndarray:
```

src/nifreeze/model/_dipy.py
```diff
@@ -41,19 +44,37 @@
...
+    """Build the GP design matrix appropriate for ``kernel``.
+
+    The b-value is always available in the Cartesian components; a single-shell
+    kernel only consumes orientations, so its trailing b-value column is dropped
+    here (multi-shell kernels keep it, selecting columns via ``bval_index``).
+    """
+    X = _cartesian_components(gtab)
+    if not isinstance(kernel, MultiShellKernel) and X.shape[-1] > 3:
+        X = X[:, :3]
+    return X
```
oesteban: Shouldn't this be done by `fit()`?

src/nifreeze/model/gpr.py
```diff
@@ -435,7 +453,7 @@ def __call__(
         if not eval_gradient:
             return self.beta_l * C_theta
 
-        # scikit-learn expects gradients w.r.t. the *log* of each hyperparameter.
+        # scikit-learn expects gradients w.r.t. the *log* of each hyperparameter
```
oesteban:
```suggestion
        # scikit-learn expects gradients w.r.t. the *log* of each hyperparameter.
```

oesteban:
I'm not positive we don't explicitly need a white kernel for the multi-shell case (#459). I also defer more thorough testing for later (e.g., #458). Code looks reasonable, merging.

PR #176, opened 2025-07-23
REF: Remove duplicate PET script

jhlegarreta:
- DOC: Fix typo in PET simulation script parameter help
- REF: Remove duplicate PET simulation data generation script


jhlegarreta:
> @mnoergaard Can you please confirm that both the removed version and the existing one provide exactly the same result that is useful for simulations? Thanks.

I have taken the time to do this; I've checked programmatically that both give the exact same result. Merging.

PR #177, opened 2025-07-23
REF: Remove PET-related output report

jhlegarreta:
Remove PET-related output report.

Added inadvertently in commit 48a876f in the context of finishing the works to add the necessary support for PET data in PR: https://github.com/nipreps/nifreeze/pull/174


jhlegarreta:
Not sure if this was some output of the notebook or something else, but I do not think it should be kept in the repository.

PR #181, opened 2025-07-24
TST: Add a fixture to create random PET data for tests

jhlegarreta:
- FIX: Use class attribute instead of outer variable in PET test
- TST: Use existing random spatial data generation fixture in PET test
- TST: Add a fixture to create random PET data for tests

### DIFF
--- a/test/test_analysis.py
+++ b/test/test_analysis.py
@@ -152,9 +152,7 @@ def test_identify_spikes(request):
     fd = rng.normal(0, 5, n_samples)
     threshold = 2.0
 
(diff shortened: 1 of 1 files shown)


jhlegarreta:
Not sure why new values are required here
https://github.com/nipreps/nifreeze/pull/181/files#diff-0a5cb82dca1264d4dbcc31121b62678b639ee753c45919b85fcbec33611d96ddR155

is necessary: the seed has not changed here, and the function does not use the fixtures. Any idea @oesteban ?

PR #185, opened 2025-07-24
REF: Generalize PET notebook data path

jhlegarreta:
Generalize PET notebook data path and make it more consistent with the rest of the notebooks.


jhlegarreta:
The simulated data will need to be uploaded to GIN so that we can test the notebook more easily.

jhlegarreta:
> The simulated data will need to be uploaded to GIN so that we can test the notebook more easily.

https://gin.g-node.org/nipreps-data/tests-nifreeze/pulls/4

jhlegarreta:
> The simulated data will need to be uploaded to GIN so that we can test the notebook more easily.

PR #199 modifies the script to download the data automatically.

PR #187, opened 2025-07-24
REF: Reuse existing FD visualization code in PET notebook

jhlegarreta:
Reuse existing FD visualization code in PET notebook.


jhlegarreta:
The look and field and layout are a different, but the function can be changed if we latter think that the one changed is better. Also, we may want to be able to change the (x-) axis name. To be done in separate PRs. Also, `plot_volumewise_motion` does not take a `DataFrame`. We can switch to a dataframe in a separate PR if that is preferred.

jhlegarreta:
This is what the plots in `main` look like:
pet_translation_rotation_mnoergaard
pet_fd_mnoergaard

This is what the plots in the proposed PR look like:
pet_translation_rotation
pet_fd

Colors, adding axis ticks and labels (for the FD plot), etc. are out of the scope of this PR.

Realized that consecutive runs do not provide the exact same output. Not sure if there is some unfixed seed here. Anyways that is out of the scope of this PR.

@mnoergaard are you OK with how the figures look like?

PR #188, opened 2025-07-24
REF: Refactor motion-related data computation functions

jhlegarreta:
Refactor motion-related data computation functions into reusable code from the PET notebook.

Rename the existing FD function to distinguish it from the refactored function.


jhlegarreta:
I'm leaving writing testing functions for a separate PR.

Issue #191, opened 2025-07-26
Refactor PET-related models


jhlegarreta:
### What would you like to see added in this software?

PR #174 finished the implementation of PET-related data, model and estimator classes. The implementations deviated from the base class implementations or did not inherit base class implementations:

- The `data.PET` class implements a `lofo_split` method. However, there is a `lovo_split` function in the `splitting` module. The data splitting function used by PET should dwell in `splitting` and the necessary refactoring/reconciling should take place so that it is sufficiently generic so as to be used by any modality.
- The `data.PET.to_nifti` function deviates from the base class signature, and thus, when adding type hinting was attempted in PR #189, `mypy` complained about this. So the signature should be reconciled with the one from the base class or the latter should be adapted.
- The statement https://github.com/nipreps/nifreeze/blob/d2def7a4fe6b7f1ab2f62120bdbb7d675e22ea70/src/nifreeze/data/pet.py#L133 seems to assume some HDF5 file structure that may not be common across modalities. This should be harmonized.
- The `PETEstimator` class generates the train/test splits internally. In general, handling the train/test splits should be done by some manager class that is common to all modalities, and thus, should be taken out of the `PETEstimator.run` function: https://github.com/nipreps/nifreeze/blob/d2def7a4fe6b7f1ab2f62120bdbb7d675e22ea70/src/nifreeze/estimator.py#L252
- The `model.pet.PETModel.fit_predict` https://github.com/nipreps/nifreeze/blob/d2def7a4fe6b7f1ab2f62120bdbb7d675e22ea70/src/nifreeze/model/pet.py#L149 said that it was taking an `integer` but it was not the case if we have a look at the PET notebook:
  ```
  predicted = model.fit_predict(pet_dataset.midframe[2])
  ```
  as suggested also by
https://github.com/nipreps/nifreeze/blob/d2def7a4fe6b7f1ab2f62120bdbb7d675e22ea70/src/nifreeze/model/pet.py#L161
and the documentation https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.BSpline.design_matrix.html

  The `index` should be an integer, as it is across other modalities. Changing the type hint causes `mypy` to complain so, there is some refactoring work that needs to be made so that `index` is an integer. X-ref https://github.com/nipreps/nifreeze/pull/189#issuecomment-3123375861.

### Do you have any interest in helping implement the feature?

Yes

jhlegarreta:
[...]

Issue #112, opened 2025-05-04
Clean up testing data and streamline downloading it across GHA workflows


jhlegarreta:
The repository used testing/example data from `OpenNeuro` and from `GIN` `G-Node`. The first are specified in a `.txt` file hosted in the latter:
https://gin.g-node.org/nipreps-data/tests-nifreeze/src/master/ismrm_sample.txt

And a script is used to download them for testing:
https://github.com/nipreps/nifreeze/blob/main/scripts/fetch_fmri_nb_openneuro_data.sh

This requires `nifreeze` to be checked out, which does not happen in the `test.yml` environment before the data caching step is performed.

At the same time, `GIN` `G-Node` hosts a file used by one of the notebooks:
https://gin.g-node.org/nipreps-data/tests-nifreeze/src/master/hcpdata.npz

However, the data in `GIN` `G-Node` is not stored as a `git-annex`.

This makes such that the steps to download and cache the data are inconsistent across GHA workflow files:
https://github.com/nipreps/nifreeze/blob/main/.github/workflows/test.yml#L36
https://github.com/nipreps/nifreeze/blob/main/.github/workflows/notebooks.yml#L44
https://github.com/nipreps/nifreeze/blob/main/.github/workflows/notebooks.yml#L54-L65
https://github.com/nipreps/nifreeze/blob/main/.github/workflows/notebooks.yml#L72

So:
- The data in `GIN` `G-Node` should be stored as a `git-annex`.
- The data downloading/caching procedures across GHA workflow files should be streamlined to reduce maintenance burden.

jhlegarreta:
Another data source to consider: PR #115 started pulling data from `OSF` to avoid excessively long download times from GIN G-Node.

PR #197, opened 2025-08-02
DOC: Document the PET notebook

jhlegarreta:
Document the PET notebook: add markdown cells explaining the steps.

### DIFF
--- a/docs/notebooks/pet_motion_estimation.ipynb
+++ b/docs/notebooks/pet_motion_estimation.ipynb
@@ -1,6 +1,27 @@
 {
  "cells": [
   {
+   "metadata": {},
+   "cell_type": "markdown",
+   "source": [
+    "# Positron emission tomography (PET) data realignment example\n",
+    "\n",
+    "This example shows how to estimate the head motion of a PET dataset using `NiFreeze`.\n",
+    "\n",
+    "The notebook uses the `sub-02` dataset that was generated synthetically from\n",
+    "a real PET dataset by adding random motion. The dataset can be installed from\n",
+    "[GIN G-node](https://gin.g-node.org/nipreps-data/tests-nifreeze):\n",
+    "\n",
+    "```\n",
+    "$ datalad install -g https://gin.g-node.org/nipreps-data/tests-nifreeze.git\n",
+    "```\n",
+    "\n",
+    "after which the environment variable `TEST_DATA_HOME` will need to be set to\n",
+    "point to the corresponding folder."
+   ],
+   "id": "6eecce4ffef2723"
+  },
+  {
    "cell_type": "code",
    "execution_count": 1,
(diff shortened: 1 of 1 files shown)


docs/notebooks/pet_motion_estimation.ipynb
```diff
@@ -1,20 +1,38 @@
...
+    "\n",
+    "This example shows how to estimate the head motion of a PET dataset using\n",
+    "`NiFreeze`.\n",
+    "\n",
+    "The notebook uses the `sub-02` dataset that was generated synthetically from\n",
+    "a real PET dataset by adding random motion. The dataset\n",
+    "can be installed from [GIN G-node](https://gin.g-node.org/nipreps-data/tests-nifreeze):\n",
+    "\n",
+    "```\n",
+    "$ datalad install -g https://gin.g-node.org/nipreps-data/tests-nifreeze.git\n",
```
jhlegarreta, line 17: Still not true as the data does not use `datalad` yet, but it will eventually (related to issue #112).

jhlegarreta:
@mnoergaard  I would be grateful if you could review this PR. There is a number of things that I do not fully understand, and may require separate work (i.e. issues and PRs)

- What is the purpose of the below statement ? The split is done internally by `PETMotionEstimator` and the `data_train`, and `data_test` are not used. Maybe it was a simple example ?
```
data_train, data_test = pet_dataset.lofo_split(15)
```

- In the statement
```
model = PETModel(dataset=pet_dataset, timepoints=pet_dataset.midframe, xlim=7000)
```

Why is `xlim` given a seemingly arbitrary value? Documenting the class may help clarifying this. Further down the notebook `pet_dataset.total_duration` is used to set the value for `xlim`.

- Is the below statement necessary?
```
model.fit_predict(None)
```

- IIUC, `midframe` is a 3D volume, so I do not fully understand what the below statement's intention is
```
predicted = model.fit_predict(pet_dataset.midframe[2])
```

I think the `predict` methods expect some sort of index (whatever its form is for the modality at issue) to estimate the data at that index. Here, `pet_dataset.midframe[2]` is something other than an index. Maybe related to issue #191.

- There is something missing in the before/after correction plot because the `PET_before` is not being shown. 

- Slightly related to the above, wouldn't it be better if we could propose (or leverage `nireports` if possible) a GIF with a flickering between uncorrected vs corrected? Otherwise the difference between the two can be hard to tell.

- Are we sure that `data_test[0]` is the data after motion correction in the following statement?
```
nifti_img_after = nb.Nifti1Image(data_test[0], pet_dataset.affine)
```

 `data_test` it is assigned by the `lofo_split` call but it is never used? Maybe the `before_mc` and `after_mc` NIfTI filenames are swapped?

- It seems weird to first make the correction with a given model instance, and then instantiate the estimator with a new model, isn't it?

PR #199, opened 2025-08-08
ENH: Fetch `sub-02` data for PET notebook

jhlegarreta:
- ENH: Use upstream fork for PET data download script
- ENH: Fetch `sub-02` data for PET notebook


jhlegarreta:
Re https://github.com/nipreps/nifreeze/pull/199#issuecomment-3169563727. This only happens if the cache is not available, which based on the recent builds does not happen that often. Thus, not a problem.

Merging.

PR #202, opened 2025-08-09
DOC: Fix grammar in comment

jhlegarreta:
Fix grammar in comment.


jhlegarreta:
No need to wait for builds in this case.

PR #203, opened 2025-08-09
REF: Refactor PET estimator

jhlegarreta:
Refactor PET estimator so that it is more closely aligned with the `Estimator` class, and in order to eventually merge them.


jhlegarreta:
I think this is close enough to the general estimator class.

I compared the current version and the one in the patch set:
- The one in the patch set is 6x faster (5:26 vs 32:10).
- Results are different, though.

Version the `PETEstimator` in `main`:

pet_estimated_motion_main

pet_fd_main

Using the `PETEstimator` in this patch set:
 
pet_estimated_motion_ps

pet_fd_ps

And none seems to be doing a good job: I have plot the ground truth:

pet_sub02_groundtruth

Given that none is doing a good job, and unless you clearly see the fix for the original `PETEstimator` @mnoergaard, my opinion is that this should be merged so that we invest our time better fixing a version that is closer to the `Estimator` class so that merging both is eventually easier.

jhlegarreta:
Closing as PR #426 has been merged.

PR #204, opened 2025-08-09
REF: Refactor PET model

jhlegarreta:
Refactor PET model: use a base class that contains the essential properties for a PET model and create a derived class that implements the B-Spline correction.

In the base class:
- Remove the `timepoints` attribute: it corresponds to the PET dataset `midframe` attribute, and thus, there is no need to duplicate it in the model.
- Remove the `xlim` attribute: it corresponds to the PET dataset `total_duration` attribute, and thus, there is no need to duplicate it in the model.
- Remove the checks related to the above attributes: checking the coherence of their values is not the model's responsibility, and should be done when instantiating the dataset.

Make the model use the frame index instead of the `midframe` value as the argument of the `fit_predict` function. This increases makes the PET model be consistent with the approach used in the code base as a principle for generalization in the framework.

The index was also marked as an integer by the type annotations (as it should be expected), whereas `midframe` values that were being provided are typically floating point values.

Change the tests and the PET notebook accordingly.

In the derived, BSpline class:
- Remove the unnecessary explicit `float` casting around the number of control points: the `dtype="float32"` specifier creates a float array.
  Fixes:
  ```
  Expected type 'str | Buffer | SupportsFloat | SupportsIndex', got 'Literal[0] | None | {__eq__} | int' instead
  ```
  raised by the IDE.

Take advantage of the commit to use `np.asarray` to avoid extra copies when not necessary.

Cast explicitly to `int` when getting the frame indices in the estimator's `run` method to avoid type checking errors:
```
src/nifreeze/estimator.py:274: error:
 Argument 1 to "fit_predict" of "BSplinePETModel" has incompatible type
 "signedinteger[_32Bit | _64Bit]"; expected "int | None"  [arg-type]
```

raised for example at:
https://github.com/nipreps/nifreeze/actions/runs/20402438629/job/58626813972#step:8:35

### DIFF
--- a/src/nifreeze/model/pet.py
+++ b/src/nifreeze/model/pet.py
@@ -22,6 +22,7 @@
 #
 """Models for nuclear imaging."""
 
+from abc import ABC, ABCMeta, abstractmethod
 from os import cpu_count
 from typing import Union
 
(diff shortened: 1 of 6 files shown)


src/nifreeze/model/pet.py
```diff
@@ -31,60 +33,86 @@
...
+        if not hasattr(dataset, "total_duration"):
+            raise TypeError("Dataset MUST be a PET object.")
+
+        if not hasattr(dataset, "midframe"):
+            raise ValueError("Dataset MUST have a midframe.")
+
+        # ToDO
+        # Are the timepoints your "gradients" ??? If so, can they be computed
+        # from frame_time or frame_duration
+        # Or else frame_time and frame_duration ????
```
jhlegarreta, line 103: @mnoergaard question for you. By your "gradients" I mean the parallel to the DWI cases, where the gradients are the independent variable where we will be willing to estimate the signal.

src/nifreeze/model/pet.py
```diff
@@ -31,60 +33,86 @@
...
+
+        self._data_mask = (
+            dataset.brainmask
+            if dataset.brainmask is not None
+            else np.ones(dataset.dataobj.shape[:3], dtype=bool)
+        )
+
+        # ToDo
+        # Are timepoints and xlim features that all PET models require ??
```
jhlegarreta, line 112: @mnoergaard question for you.

src/nifreeze/model/pet.py
```diff
@@ -111,17 +132,24 @@ def is_fitted(self):
...
+        n_jobs = n_jobs or 1
+
         if self._locked_fit is not None:
             return n_jobs
 
         if index is not None:
             raise NotImplementedError("Fitting with held-out data is not supported")
+
+        # ToDo
+        # Does not make sense to make timepoints be a kwarg if it is provided as a named parameter to __init__
```
jhlegarreta, line 144: @mnoergaard question for you.

src/nifreeze/model/pet.py
```diff
@@ -135,6 +163,135 @@ def _fit(self, index: int | None = None, n_jobs=None, **kwargs):
...
+
+        if self._locked_fit is not None:
+            return n_jobs
+
+        if index is not None:
+            raise NotImplementedError("Fitting with held-out data is not supported")
+
+        # ToDo
+        # Does not make sense to make timepoints be a kwarg if it is provided as a named parameter to __init__
+        timepoints = kwargs.get("timepoints", None) or self._x
```
jhlegarreta, line 290: Same as https://github.com/nipreps/nifreeze/pull/204/files#r2271605506.

src/nifreeze/model/pet.py
```diff
@@ -149,6 +306,12 @@ def _fit(self, index: int | None = None, n_jobs=None, **kwargs):
     def fit_predict(self, index: int | None = None, **kwargs):
         """Return the corrected volume using B-spline interpolation."""
 
+        # ToDo
+        # Does the below apply to PET ? Martin has the return None statement
+        # if index is None:
+        #    raise RuntimeError(
+        #        f"Model {self.__class__.__name__} does not allow locking.")
```
jhlegarreta, line 313: Another question for @mnoergaard.

src/nifreeze/model/pet.py
mnoergaard: Timepoints for PET are the gradients of DWI. We subsample the PET data into frames, and frames have a start time and an end time, and the values from the PET data are the uptake within this time period. Therefore, we use the middle timepoint (midframe) as timepoint.

src/nifreeze/model/pet.py
```diff
@@ -95,62 +125,114 @@ def __init__(
         if self._x[-1] > (self._xlim - DEFAULT_TIMEFRAME_MIDPOINT_TOL):
             raise ValueError("Last frame midpoint should not be equal or greater than duration")
 
-        # Calculate index coordinates in the B-Spline grid
-        self._n_ctrl = n_ctrl or (len(timepoints) // 4) + 1
+    def _preproces_data(self) -> np.ndarray:
```
mnoergaard, line 128: preprocess?

src/nifreeze/model/pet.py
```diff
@@ -95,62 +125,114 @@ def __init__(
...
-            )
-            data = smoothed_img.get_fdata()
-
-        if self._thresh_pct > 0:
-            thresh_val = np.percentile(data, self._thresh_pct)
-            data[data < thresh_val] = 0
-
-        # Convert data into V (voxels) x T (timepoints)
-        data = data.reshape((-1, data.shape[-1])) if brainmask is None else data[brainmask]
+        # ToDo
```
mnoergaard, line 209: Allowing _fit to override timepoints through kwargs duplicates information that was already validated and stored on self._x during initialization. Dropping that extra kwarg simplifies the API and avoids inconsistent inputs

src/nifreeze/model/pet.py
```diff
@@ -95,62 +125,114 @@ def __init__(
...
+        # Does not make sense to make timepoints be a kwarg if it is provided as a named parameter to __init__
+        timepoints = kwargs.get("timepoints", None) or self._x
```
mnoergaard, line 211: should be removed

src/nifreeze/model/pet.py
jhlegarreta, line 128:
```suggestion
    def _preprocess_data(self) -> np.ndarray:
```

src/nifreeze/model/pet.py
```diff
@@ -95,62 +125,114 @@ def __init__(
...
             return n_jobs
 
         if index is not None:
             raise NotImplementedError("Fitting with held-out data is not supported")
-        timepoints = kwargs.get("timepoints", None) or self._x
-        x = (np.array(timepoints, dtype="float32") / self._xlim) * self._n_ctrl
 
-        data = self._dataset.dataobj
-        brainmask = self._dataset.brainmask
+        data = self._preproces_data()
```
jhlegarreta, line 207:
```suggestion
        data = self._preprocess_data()
```

src/nifreeze/model/pet.py
jhlegarreta: Resolved in b2898e2. I will squash once the rest gets resolved.

src/nifreeze/model/pet.py
jhlegarreta: OK, understood. Thanks. Removing the comment and resolving.

src/nifreeze/model/pet.py
```diff
@@ -95,62 +125,114 @@ def __init__(
         if self._x[-1] > (self._xlim - DEFAULT_TIMEFRAME_MIDPOINT_TOL):
             raise ValueError("Last frame midpoint should not be equal or greater than duration")
 
-        # Calculate index coordinates in the B-Spline grid
-        self._n_ctrl = n_ctrl or (len(timepoints) // 4) + 1
+    def _preproces_data(self) -> np.ndarray:
+        # ToDo
```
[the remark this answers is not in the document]
jhlegarreta: If I recall correctly, the idea there is to preprocess only the data that is required; e.g. if we are using only a subset of the volumes, then we shouldn't preprocess the entire data, e.g. 
https://github.com/nipreps/nifreeze/blob/4cfaadb96ea9f7a3a31797c76d931d770a661ded/src/nifreeze/model/dmri.py#L122-L124

The `data, _, gtab = self._dataset[idxmask]` is commented out in here immediately after the `ToDo` to as a pointer to that idea.

src/nifreeze/model/pet.py
jhlegarreta: Not sure I understand: maybe the reply is answering a question elsewhere?

PR #208, opened 2025-08-12
STY: Re-use and define global variables for dMRI model

jhlegarreta:
- STY: Re-use the lower-bound b and S0 thresholds from dMRI data module
- STY: Define the S0 clipping percentile as a global variable

### DIFF
--- a/src/nifreeze/model/dmri.py
+++ b/src/nifreeze/model/dmri.py
@@ -28,12 +28,9 @@ import numpy as np
 from dipy.core.gradients import gradient_table_from_bvals_bvecs
 from joblib import Parallel, delayed
 
-from nifreeze.data.dmri import DTI_MIN_ORIENTATIONS, DWI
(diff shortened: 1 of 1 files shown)


src/nifreeze/model/dmri.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, dataset: DWI, max_b: float | int | None = None, **kwargs):
                 f"DWI dataset is too small ({dataset.gradients.shape[0]} directions)."
             )
 
-        if max_b is not None and max_b > B_MIN:
+        if max_b is not None and max_b > DEFAULT_LOWB_THRESHOLD:
```
jhlegarreta, line 82: Note that as explained in https://github.com/nipreps/nifreeze/pull/208/commits/cd9f6cabed30e8119480c47b2da1231b55d14361, the S0 lower bound is changed from 1e-6 to 1e-5.

PR #210, opened 2025-08-13
FIX: Fix miscellaneous oversights in the base DWI model

jhlegarreta:
Fix miscellaneous oversights in the base DWI model:
- Provide the `None` default value for the `n_jobs` kwarg if not provided in the dictionary.

  Fixes:
  ```
          n_models = self._fit(
              index,
  >           n_jobs=kwargs.pop("n_jobs"),
              **kwargs,
          )
  E       KeyError: 'n_jobs'
  ```

- Make the diffusion gradients have a unit norm in the gtab generation fixture.

  Fixes:
  ```
          bvecs = np.where(np.isnan(bvecs), 0, bvecs)
          bvecs_close_to_1 = abs(vector_norm(bvecs) - 1) <= atol
          if bvecs.shape[1] != 3:
              raise ValueError("bvecs should be (N, 3)")
          if not np.all(bvecs_close_to_1[dwi_mask]):
  >           raise ValueError(
                  "The vectors in bvecs should be unit (The tolerance "
                  "can be modified as an input parameter)"
              )
  E           ValueError: The vectors in bvecs should be unit (The tolerance can be modified as an input parameter)
  ```

Add a test to check that DIPY models (e.g. DTI) can be instantiated and that the returned prediction has the expected shape.

### DIFF
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -157,6 +157,31 @@ def test_gp_model(evals, S0, snr, hsph_dirs, bval_shell):
     assert prediction.shape == (2,)
 
 
+@pytest.mark.random_dwi_data(50, (14, 16, 8), True)
+def test_dti_model(setup_random_dwi_data):
+    (
+        dwi_dataobj,
+        affine,
+        brainmask_dataobj,
+        b0_dataobj,
+        gradients,
+        _,
+    ) = setup_random_dwi_data
+
+    dataset = DWI(
+        dataobj=dwi_dataobj,
+        affine=affine,
+        brainmask=brainmask_dataobj,
+        bzero=b0_dataobj,
+        gradients=gradients,
+    )
+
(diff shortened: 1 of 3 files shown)


test/test_model.py
```diff
@@ -157,6 +157,33 @@ def test_gp_model(evals, S0, snr, hsph_dirs, bval_shell):
...
+        affine=affine,
+        brainmask=brainmask_dataobj,
+        bzero=b0_dataobj,
+        gradients=gradients
+    )
+
+    dtimodel = model.DTIModel(dataset)
+    predicted = dtimodel.fit_predict(4)
+
+    assert predicted.shape == dwi_dataobj.shape[:-1]
```
jhlegarreta, line 184: I assume we trust the prediction makes sense and that checking the shape here is enough.

PR #225, opened 2025-09-15
DOC: Remove legacy eddymotion DOI from README file

jhlegarreta:
Remove legacy eddymotion DOI from README file: the package has been revamped significantly since the work referenced by that DOI, has been renamed to `nifreeze` and can be cited through the reference contained in the `CITATION.cff` file.


jhlegarreta:
`Benchmark` build failure is unrelated. Will be investigated in a separate PR.

PR #227, opened 2025-09-15
DOC: Update the toolkit usage documentation

### LINKED ISSUE #224: Estimator usage documentation is outdaded/not sync'ed with the implementation
### What would you like changed/added and why?

A user reported offline that the usage documentation is outdated: in the provided example `Estimator()` doesn't take any parameters
https://github.com/nipreps/nifreeze/blob/18bec0ddf399bf50d488ec841f7b674bf320945c/docs/usage.rst?plain=1#L40

and the model is specified in `estimate()`:
https://github.com/nipreps/nifreeze/blob/18bec0ddf399bf50d488ec841f7b674bf320945c/docs/usage.rst?plain=1#L49

However, the current implementation of the command line tool is different, as the `Estimator` instantiation takes some parameters, including the model:
https://github.com/nipreps/nifreeze/blob/18bec0ddf399bf50d488ec841f7b674bf320945c/src/nifreeze/estimator.py#L84-L92

Using the implementation of the command line tool in the source allows the pipeline to effectively run.

The documentation should be updated.

### Do you have any interest in helping improve the documentation?

Yes

### Do you have any suggestions for the new documents?

_No response_

jhlegarreta:
Update the toolkit usage documentation: PR https://github.com/nipreps/nifreeze/pull/62 introduced important changes to the approach that had been used so far within the toolkit. That PR missed to update the usage documentation.

This patch set:
- Expands the first section on importing the `NiFreeze` components to explain at a high-level the components required to run the `NiFreeze` motion estimation.
- Generalizes all passages and make them agnostic to the modality being used as much as possible.
- Updates the instantiation of the `Estimator`, the call to the actual estimation method (`run` instead of `estimate` after the changes introduced in the mentioned PR), and the change in the approach to store the estimated motion parameters (now held by the dataset class instead of being returned by the estimator) to the current API.
- Explains the parameters of the estimator.
- Updates the plotting section to include the motion parameter visualization functions, and refers to `NiReports` for other functions that were outsourced to there.
- Uses hyperlinks to the API documentation (modules, classes, etc.) where possible.
- Uses the inline `:math:` mode where appropriate.

Take advantage of the commit to:
- Use double backticks for inline verbatim code as dictated by Sphinx.
- Italicize `NiFreeze`.
- Use double quotes to match the style used in the code.

Fixes #224.

### DIFF
--- a/docs/links.rst
+++ b/docs/links.rst
@@ -1,4 +1,5 @@
 .. _Nipype: http://nipype.readthedocs.io/en/latest/
+.. _NiReports: https://nireports.readthedocs.io/en/latest/
 .. _BIDS: http://bids.neuroimaging.io
 .. _Installation: installation.html
 .. _workflows: workflows.html
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -4,9 +4,23 @@ How to Use
 ==========
 Incorporating NiFreeze into a Python module or script
 -----------------------------------------------------
-To utilize NiFreeze functionalities within your Python module or script, follow these steps:
+To utilize *NiFreeze* functionalities within your Python module or script, follow these steps:
 
-1. **Import NiFreeze Components**: Start by importing necessary components from the NiFreeze package:
+1. **Import NiFreeze Components**: Start by importing necessary components from the *NiFreeze* package:
+
+   To use *NiFreeze* you will typically need to use three main components:
+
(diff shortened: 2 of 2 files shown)


jhlegarreta:
Benchmark build failure is unrelated. Will be addressed in a separate PR.

PR #231, opened 2025-09-24
ENH: Add uptake statistic property to PET data

jhlegarreta:
Add uptake statistic property to PET data: compute the data as the sum of voxel values across a frame.


jhlegarreta:
This will be useful when iterating over uptake values: x-ref PR #228.

PR #235, opened 2025-09-26
STY: Fix `NiTransforms` classname annotation

jhlegarreta:
Fix `NiTransforms` classname annotation: use `TransformBase` instead of the non-existing `BaseTransform`.


jhlegarreta:
What `nitransform` shows as of today as the classname:

Screenshot from 2025-09-26 19-32-09

Issue #230, opened 2025-09-23
Generalize the linear iterator to handle data other than DWI data


jhlegarreta:
### What would you like to see added in this software?

The `linear_iterator` can handle only DWI data, as it requires `bvals` to be present in the kwargs list:
https://github.com/nipreps/nifreeze/blob/a4edd6219f7288410d6895cf8f23fe846776b5bc/src/nifreeze/utils/iterators.py#L30

It should be generalized to be able to traverse data other than DWI data.

So what are the features that we will be willing to traverse for:
- fMRI data?
- PET data?

### Do you have any interest in helping implement the feature?

Yes

jhlegarreta:
Generally, all current iterators, not only the linear iterator, expect to find the `bval` property. Thus, the `bvalue_iterator` and `uptake_iterator` (the latter introduced in PR #228) naming may be misleading. Overall, any iterator should accept the property of interest for any modality of interest, and a single function doing the job for all of them.

Now, we may need to think of renaming `bvalue_iterator` and `uptake_iterator` to something more general. Note that this is an API break.

PR #238, opened 2025-09-27
ENH: Generalize iterators to accept `bvals` or `uptake` as kwargs

jhlegarreta:
Generalize iterators to accept `bvals` or `uptake` as keyword arguments. The previous implementation of the linear, random and centralsym iterators was only accepting `bvals`. This patch set allows these iterators to work with PET data through the `uptake` argument.

Transition to keyword argument-only style. Adapt the doctests accordingly.

Document the functions by explicitly assigning the docstring to the `__doc__` property of each function so that the `SIZE_KEYS_DOC` can be reused and to allow the examples be run by the doctring tests.

### DIFF
--- a/src/nifreeze/utils/iterators.py
+++ b/src/nifreeze/utils/iterators.py
@@ -26,7 +26,24 @@ import random
 from itertools import chain, zip_longest
 from typing import Iterator
 
-ITERATOR_SIZE_ERROR_MSG = "Cannot build iterator without size."
+SIZE_KEYS = ("size", "bvals", "uptake")
+"""Keys that may be used to infer the number of volumes in a dataset. When the
+size of the structure to iterate over is not given explicitly, these keys
+correspond to properties that distinguish one imaging modality from another, and
+are part of the 4th axis (e.g. diffusion gradients in DWI or update in PET)."""
+
+SIZE_KEYS_DOC = """
+size : obj:`int`, optional
+    Size of the structure to iterate over.
+bvals : :obj:`list`, optional
+    List of b-values corresponding to all orientations of a DWI dataset.
+uptake : :obj:`list`, optional
+    List of uptake values corresponding to all volumes of the dataset.
+"""
+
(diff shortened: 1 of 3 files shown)


jhlegarreta:
Partially addressed #230. It remains the question about the `bvalue` and `uptake` iterators: https://github.com/nipreps/nifreeze/issues/230#issuecomment-3341877860.

src/nifreeze/utils/iterators.py
```diff
@@ -26,91 +26,93 @@
...
-    --------
-    >>> list(linear_iterator(10))
-    [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+    present_keys = [k for k in SIZE_KEYS if k in kwargs]
+    if len(present_keys) > 1:
+        raise ValueError(f"Only one of {SIZE_KEYS} may be provided (got: {present_keys})")
+    if present_keys:
+        candidate = kwargs[present_keys[0]]
+        return candidate if isinstance(candidate, int) else len(candidate)
+    return None
```
oesteban:
```suggestion
    candidates = [kwargs[k] for k in SIZE_KEYS if k in kwargs]
    if candidates:
        return candidates[0] if isinstance(candidates[0], int) else len(candidates[0])
    raise ValueError(f"None of {SIZE_KEYS} were provided to infer size")
```

src/nifreeze/utils/iterators.py
```diff
@@ -36,92 +51,74 @@
...
 
-    Other Parameters
-    ----------------
-    bvals : :obj:`list`
-        List of b-values corresponding to all orientations of a DWI dataset.
-        If ``size`` is provided, this argument will be ignored.
-        Otherwise, ``size`` will be inferred from the length of ``bvals``.
+    Returns
+    -------
+    :obj:`int` or None
```
oesteban:
```suggestion
    :obj:`int`
```

PR #243, opened 2025-09-29
TST: Test iterators

jhlegarreta:
Test iterators. Even if docstring tests cover the core fetures, exceptions remain untested. This patch set adds unit tests that do test the exceptions and the different use cases with the keyword arguments that iterators accept.

Take advantage of the commit to define the keyword arguments and exception messages as global variables so that the exceptions raised by the iterators get checked.

### DIFF
--- a/src/nifreeze/utils/iterators.py
+++ b/src/nifreeze/utils/iterators.py
@@ -26,6 +26,15 @@ import random
 from itertools import chain, zip_longest
 from typing import Iterator
 
+ITERATOR_SIZE_ERROR_MSG = "Cannot build iterator without size."
+"""Iterator size argument error message."""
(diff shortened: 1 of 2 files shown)


jhlegarreta:
Do we prefer `kwargs.pop` so that we can accept round_decimals and pass them to `_value_iterator`?
https://github.com/nipreps/nifreeze/blob/4a96c3cf1a22b124918db53777f868bbd5fbfbc0/src/nifreeze/utils/iterators.py#L187
https://github.com/nipreps/nifreeze/blob/4a96c3cf1a22b124918db53777f868bbd5fbfbc0/src/nifreeze/utils/iterators.py#L219

Will add the `uptake` kwarg tests once PR #238 gets merged.

oesteban (approved):

PR #247, opened 2025-09-30
ENH: Allow specifying decimal places in value iterators

jhlegarreta:
Allow specifying decimal places in value iterators: by popping the name of the kwarg name of interest instead of getting it, kwargs (which can contain the `decimal_places` parameter) can be passed to the private function accepting the value of decimal places.


jhlegarreta:
PR #243 can be merged as is, and would be sensible to merge it first so that the iterators get proper testing. This PR and PR #238 would then need to be rebased on `main` so that the necessary aspects are adapted, e.g. https://github.com/nipreps/nifreeze/pull/243#issuecomment-3349394871

PR #250, opened 2025-10-01
FIX: Allow instantiating the PET model through the factory

jhlegarreta:
Allow instantiating the PET model through the factory.

### DIFF
--- a/src/nifreeze/model/base.py
+++ b/src/nifreeze/model/base.py
@@ -72,8 +72,10 @@ class ModelFactory:
         if model.lower() in ("gqi", "dti", "dki", "pet"):
             from importlib import import_module
 
-            dmrimod = import_module("nifreeze.model.dmri")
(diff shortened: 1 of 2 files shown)


src/nifreeze/model/base.py
```diff
@@ -69,7 +69,7 @@ def init(model: str | None = None, **kwargs):
 
             return AverageDWIModel(kwargs.pop("dataset"), **kwargs)
 
-        if model.lower() in ("gqi", "dti", "dki", "pet"):
+        if model.lower() in ("gqi", "dti", "dki"):
             from importlib import import_module
 
             dmrimod = import_module("nifreeze.model.dmri")
```
oesteban: Would it make more sense to keep the "pet" keyword and make this more generic?

```suggestion
            thismod = import_module(f"nifreeze.model.{'pet' if model.lower() == 'pet' else 'dmri'}")
```

PR #252, opened 2025-10-01
TST: Test dMRI model instantiation through factory

jhlegarreta:
Test instantiating the dMRI models through the factory.


jhlegarreta:
If the PET models should also be allowed to be instantiated through the factory (related PR #250) the iterator, this will be added and modified in a separate PR as the factory will require fixes. Merging.

PR #257, opened 2025-10-04
TST: Test model instantiation by estimator given name

jhlegarreta:
Test that the estimator handles correctly instantiating models through the factory when given a model name.

Make the dummy model return a random array instead of an all-ones array as its prediction to avoid the advanced clip function raise invalid value warnings when called by the NIfTI serialization functions of the registration data preparation process. Avoids:
```
test/test_estimator.py::test_estimator_init_model_string
    /home/runner/work/nifreeze/nifreeze/src/nifreeze/data/filtering.py:107:
 RuntimeWarning: invalid value encountered in divide
      data /= data.max()

  test/test_estimator.py::test_estimator_init_model_string
    /home/runner/work/nifreeze/nifreeze/src/nifreeze/data/filtering.py:113:
 RuntimeWarning: invalid value encountered in cast
      np.round(255 * data, out=data).astype(dtype)
```

raised for example at:
https://github.com/nipreps/nifreeze/actions/runs/18264648044/job/51997242088#step:11:3905

### DIFF
--- a/test/test_estimator.py
+++ b/test/test_estimator.py
@@ -31,9 +31,13 @@ DATAOBJ_SIZE = (5, 5, 5, 4)
 
 
 class DummyModel(BaseModel):
+    def __init__(self, **kwargs):
+        self._rng = np.random.default_rng(1234)
+        super().__init__(**kwargs)
+
     def fit_predict(self, idx=None, **kwargs):
         # Return a synthetic 3D image
-        return np.ones(DATAOBJ_SIZE[:-1])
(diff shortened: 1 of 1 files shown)


jhlegarreta:
When all values of the array were ones, the normalization was making such that all values were zero, raising an invalid-valued division:
https://github.com/nipreps/nifreeze/blob/5b7fb3a26d4e2eca9facb14b0b7ceba91f227023/src/nifreeze/data/filtering.py#L106-L107

PR #260, opened 2025-10-04
REF: Remove checks as PET function frame duration is not optional

jhlegarreta:
Remove unnecessary checks as PET function frame duration is not optional.

Adjust the docstring accordingly.


src/nifreeze/data/pet.py
```diff
@@ -277,21 +274,18 @@ def from_nii(
 
     pet_obj.uptake = _compute_uptake_statistic(data, stat_func=np.sum)
 
-    # If the user supplied new values, set them
-    if frame_time is not None:
```
jhlegarreta, line 281: At this point `pet_obj.mdiframe is `None`, so there is no "new" values.

PR #264, opened 2025-10-05
TST: Check that the general loader can load a PET HDF5 dataset

jhlegarreta:
Check that the general loader can load a PET HDF5 dataset.


jhlegarreta:
Requires a dataset file. Pinging @mnoergaard.

PR #290, opened 2025-10-25
REF: Remove unnecessary PET data class `midframe` checks for `None`

jhlegarreta:
Remove unnecessary PET data class `midframe` attribute checks for `None` values: `midframe` is a necessary piece of information for the class and thus, cannot be `None`: when reading a PET sequence, it is ultimately set at
https://github.com/nipreps/nifreeze/blob/eebdceb0ead93abab6a68685a684ece92cdfc526/src/nifreeze/data/pet.py#L290


jhlegarreta:
Similar to PR #289.

PR #294, opened 2025-10-25
STY: Mark the base data class `brainmask` attribute as optional

jhlegarreta:
Mark the base data class `brainmask` attribute as optional: add `None` to the possible types it can have.

Fixes:
```
test/test_model.py:91: error:
 Argument "brainmask" to "DWI" has incompatible type "ndarray[Any, dtype[Any]] | None";
 expected "ndarray[Any, Any]"  [arg-type]
```

Assign `None` when creating the brainmask data in the `motion_data` fixture in case the data stored in the testing HDF5 file has no brainmask data. Avoids:
```
test/conftest.py:95: error:
 Item "None" of "ndarray[Any, Any] | None" has no attribute "astype"  [union-attr]
```

Ensure that the brainmask is not `None` through an assertion rather than checking whether the PET object has the `brainmask` attribute. Avoids:
```
test/test_data_pet.py:145: error:
 Argument 1 to "assert_array_equal" has incompatible type "ndarray[Any, Any] | None";
 expected "_SupportsArray[dtype[Any]] | _NestedSequence[_SupportsArray[dtype[Any]]] |
 bool | int | float | complex | str | bytes | _NestedSequence[bool | int | float |
 complex | str | bytes]"  [arg-type]
```

raised locally after adding `None` to the `brainmask` type annotation.

### DIFF
--- a/src/nifreeze/data/base.py
+++ b/src/nifreeze/data/base.py
@@ -78,7 +78,9 @@ class BaseDataset(Generic[Unpack[Ts]]):
     """A :obj:`~numpy.ndarray` object for the data array."""
     affine: np.ndarray = attrs.field(default=None, repr=_data_repr, eq=attrs.cmp_using(eq=_cmp))
     """Best affine for RAS-to-voxel conversion of coordinates (NIfTI header)."""
-    brainmask: np.ndarray = attrs.field(default=None, repr=_data_repr, eq=attrs.cmp_using(eq=_cmp))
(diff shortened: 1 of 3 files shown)


jhlegarreta:
Fixes one out of the three remaining errors in PR #273:
https://github.com/nipreps/nifreeze/actions/runs/18478441375/job/52648138075?pr=273#step:8:34

PR #295, opened 2025-10-25
STY: Allow `fit_predict` to return `None`

jhlegarreta:
Allow `fit_predict` to return `None` as the `_locked_fit` attribute can be `None`: add `None` to the return type annotations of the method.

Fixes:
```
src/nifreeze/model/base.py:147: error:
 Incompatible return value type (got "Any | None",
 expected "ndarray[Any, Any]")  [return-value]
```

raised for example at:
https://github.com/nipreps/nifreeze/actions/runs/18478441375/job/52648138075?pr=273#step:8:35

### DIFF
--- a/src/nifreeze/model/base.py
+++ b/src/nifreeze/model/base.py
@@ -23,7 +23,6 @@
 """Base infrastructure for nifreeze's models."""
 
 from abc import ABC, ABCMeta, abstractmethod
-from typing import Union
 from warnings import warn
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


src/nifreeze/model/base.py
```diff
@@ -140,7 +140,7 @@ def __init__(self, dataset, predicted=None, **kwargs):
         if self._locked_fit is None:
             raise TypeError("This model requires the predicted map at initialization")
 
-    def fit_predict(self, *_, **kwargs) -> np.ndarray:
+    def fit_predict(self, *_, **kwargs) -> np.ndarray | None:
```
oesteban: This can't be None for this particular model... That said, it'd be better to fix the type annotation on line 107 for consistency.

PR #296, opened 2025-10-25
REF: Refactor the `get_extra` attribute to return an array or `None`

jhlegarreta:
Refactor the `get_extra` attribute to return an array or `None` instead of a tuple: the derivative classes can only return an array as the values they return can only be arrays, and the base class must return `None` to remain generic and as it does not store any extra property.

Thus, there is no need to unpack the extras in the base class `__getitem__` method.

### DIFF
--- a/src/nifreeze/data/base.py
+++ b/src/nifreeze/data/base.py
@@ -99,12 +99,12 @@ class BaseDataset(Generic[Unpack[Ts]]):
 
         return self.dataobj.shape[-1]
 
-    def _getextra(self, idx: int | slice | tuple | np.ndarray) -> tuple[Unpack[Ts]]:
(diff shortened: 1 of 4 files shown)


src/nifreeze/data/base.py
```diff
@@ -99,12 +99,12 @@ def __len__(self) -> int:
 
         return self.dataobj.shape[-1]
 
-    def _getextra(self, idx: int | slice | tuple | np.ndarray) -> tuple[Unpack[Ts]]:
-        return ()  # type: ignore[return-value]
+    def _getextra(self, idx: int | slice | tuple | np.ndarray) -> np.ndarray | None:
```
oesteban: Why changing the tuple? The idea here was that we can add extra objects if necessary in the future

PR #303, opened 2025-11-01
ENH: Generalize bvals and uptake value iterators into a single iterator

### LINKED ISSUE #230: Generalize the linear iterator to handle data other than DWI data

jhlegarreta:
Generalize bvals and uptake value iterators into a single iterator that is able to traverse the values in ascending or descending order depending on whether it is provided a list of b-values (DWI) or uptake values (PET).

Follow-up to commit a1310e6.

Fixes #230.

### DIFF
--- a/src/nifreeze/utils/iterators.py
+++ b/src/nifreeze/utils/iterators.py
@@ -24,7 +24,7 @@
 
 import random
 from itertools import chain, zip_longest
-from typing import Iterator
+from typing import Iterator, Sequence
 
 DEFAULT_ROUND_DECIMALS = 2
 """Round decimals to use when comparing values to be sorted for iteration purposes."""
@@ -172,14 +172,14 @@ Examples
 
 
 def _value_iterator(
-    values: list, ascending: bool, round_decimals: int = DEFAULT_ROUND_DECIMALS
+    values: Sequence[float], ascending: bool, round_decimals: int = DEFAULT_ROUND_DECIMALS
 ) -> Iterator[int]:
     """
     Traverse the given values in ascending or descenting order.
 
     Parameters
     ----------
-    values : :obj:`list`
+    values : :obj:`Sequence`
         List of values to traverse.
     ascending : :obj:`bool`
         If ``True``, traverse in ascending order; traverse in descending order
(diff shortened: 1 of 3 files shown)


jhlegarreta:
Sorry I cannot make heads or tails of the error:
```
 traceback_args=("No such file or directory '['/tmp/tmpn3qawmpk/ants-000030GenericAffine.mat']' for output
 'reverse_forward_transforms' of a Registration interface\nAn exception of type FileNotFoundError occurred
 while running interface Registration.",), version='2.5.4')

(...)

  E       traits.trait_errors.TraitError: Each element of the 'reverse_forward_transforms' trait of a
 RegistrationOutputSpec instance must be a pathlike object or string representing an existing file,
 but a value of '/tmp/tmpn3qawmpk/ants-000030GenericAffine.mat' <class 'str'> was specified.

(...)

  FAILED test/test_estimator.py::test_estimator_init_model_string - FileNotFoundError: No such file or directory '['/tmp/tmpn3qawmpk/ants-000030GenericAffine.mat']' for output 'reverse_forward_transforms' of a Registration interface
```

`main` has started failing since yesterday as well with the same error:
https://github.com/nipreps/nifreeze/actions/runs/18957141877/job/54136495795
vs last successful execution after a PR had been merged:
https://github.com/nipreps/nifreeze/actions/runs/18956164364/job/54133423410

We have not made changes to any components related to the transforms, `nitransforms` remains pinned and the `nipype` version used has remained the same across the above builds. The only notable difference I can see is that ITK 5.4.4 is being installed along with ANTs instead of ITK 5.4.0.

The PR immediately preceding the failure succeeded with ITK 5.4.0:
https://github.com/nipreps/nifreeze/actions/runs/18956874170/job/54135672434

oesteban:
> ITK 5.4.4 is being installed along with ANTs instead of ITK 5.4.0.

Indeed, it looks like ANTs is on the same version. That's ver weird.



> One thing that changes across failures is the name or iteration of the registration at which the test fails

This makes sense if we are using the random iterator without a seed (which we are possibly doing?) We should fix a seed so that we iterate the dataset consistently across builds.

If we are setting a seed, the seeding may not be properly working.

oesteban (approved):

PR #304, opened 2025-11-01
STY: Annotate DWI optional data class parameters as such

jhlegarreta:
Annotate DWI optional data class parameters as such: add `None` to the type annotation of `bzero` and `eddy_xfms` attributes.

Ignore the `bzero` indexing static checking error when serializing the data to NIfTI: the indexing operation is protected by a conditional `no_bzero` statement, so at that point the class instance is bound to have a non-`None` `bzero` attribute value.

Assert that `bzero` is not `None` in tests to avoid static type checking errors.

### DIFF
--- a/test/test_data_dmri.py
+++ b/test/test_data_dmri.py
@@ -122,6 +122,8 @@ def test_load(datadir, tmp_path, insert_b0, rotate_bvecs):  # noqa: C901
         )
 
     if insert_b0:
+        assert dwi_h5.bzero is not None
+        assert dwi_from_nifti1.bzero is not None
         assert np.allclose(dwi_h5.bzero, dwi_from_nifti1.bzero)
 
     assert np.allclose(dwi_h5.bvals, dwi_from_nifti1.bvals, atol=1e-3)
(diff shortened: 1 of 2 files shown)


test/test_data_dmri.py
```diff
@@ -122,6 +122,8 @@ def test_load(datadir, tmp_path, insert_b0, rotate_bvecs):  # noqa: C901
         )
 
     if insert_b0:
+        assert dwi_h5.bzero is not None
```
oesteban: We don't seem to test `dwi_h5.bzero is None` anywhere, right? Would it make sense to add one test?

jhlegarreta: For context, the assertion has been added to avoid a mypy error: https://github.com/nipreps/nifreeze/pull/304#issue-3577707146

> We don't seem to test dwi_h5.bzero is None anywhere, right? 

I do not think so. In general, loading and instantiating data classes with no/part of/all optional parameters needs to be tested more thoroughly. See below.

> Would it make sense to add one test?

X-ref https://github.com/nipreps/nifreeze/pull/304#issuecomment-3476104999. Something like this could be done to check all possible execution paths: https://github.com/nipreps/nifreeze/pull/313/files. Adding any test is out of the scope of this PR (as long as CIs do not fail), though, so please open an issue to add such a test if you do not have the bandwidth now.

PR #308, opened 2025-11-02
TST: Fix dummy DWI and PET dataset `__getitem__` method return values

jhlegarreta:
Fix estimator test dummy DWI and PET dataset `__getitem__` method return values in:
- Return `None` instead of the brainmasks as the values of the motion affines.
- Return the gradient data corresponding to the requested index for the DWI dataset.

### DIFF
--- a/test/test_estimator.py
+++ b/test/test_estimator.py
@@ -75,7 +75,7 @@ class DummyDWIDataset(BaseDataset):
         return self.dataobj.shape[-1]
 
     def __getitem__(self, idx):
-        return self.dataobj[..., idx], self.brainmask, self.gradients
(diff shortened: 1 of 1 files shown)


oesteban:
> Failures are unrelated. Merging.

Sure, but we should refrain from merging anything until the reason why the tests started to fail is figured out. It looks like ants may have changed the interface and we should either pin ants or add a counter-measure.

However, if we keep merging PRs with failing tests we may be introducing regressions that will not be evident until the root-cause of this failure is found (and, compounding errors makes it harder to resolve).

Opened #309. Let's get main to a sane state and then keep going.

PR #312, opened 2025-11-08
TST: Make estimator test dummy model class be deterministic

jhlegarreta:
Make the model class deterministic: make it return the indexed volume instead of random data.

Take advantage to:
- Make the boolean mask have some `False` values to make the dataset have some more variability, closer to a real case.
- Reword some comments for the sake of consistency.

### DIFF
--- a/test/test_estimator.py
+++ b/test/test_estimator.py
@@ -36,27 +36,26 @@ from nifreeze.utils import iterators
 DATAOBJ_SIZE = (5, 5, 5, 4)
 
 
-class DummyModel(BaseModel):
-    def __init__(self, **kwargs):
-        self._rng = np.random.default_rng(1234)
-        super().__init__(**kwargs)
+class DummyInsiderModel(BaseModel):
+    def __init__(self, dataset, **kwargs):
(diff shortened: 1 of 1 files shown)


jhlegarreta:
Well, this failed with yet something else:
```
FAILED test/test_integration.py::test_proximity_estimator_trivial_model -
 AssertionError: assert 3.31941139023541 < 0.25
```

at https://github.com/nipreps/nifreeze/actions/runs/19189601911/job/54862063974?pr=312#step:11:3063

This is the same error that we are having intermittently in some of the last PRs, e.g. most recently at
https://github.com/nipreps/nifreeze/actions/runs/19082975255/job/54516259421?pr=262#step:11:3067

in PR #262. This is different from the error that triggered this PR: https://github.com/nipreps/nifreeze/pull/257#discussion_r2506341054

Our CIs stop when they encounter the first error :-S, and given that there are multiple threads executing tests, we may be tempted to think that in reality we still have both errors, but the CI here shows:
```
  [gw1] PASSED test/test_estimator.py::test_estimator_init_model_string 
```

PR #319, opened 2025-11-09
REF: Fix PET model exception test functions

jhlegarreta:
Fix PET model exception test functions:
- Refactor functions so that the last midpoint value exception is properly captured: rename tests functions so that their purposes becomes apparent from the name and add a test to check the last midpoint value exception.
- Check both `midframe` and `xlim` missing exceptions.
- Define the exception messages so that tests can check exactly the captured message.


jhlegarreta:
Finally done in PR #318. Closing.

PR #320, opened 2025-11-11
FIX: Flaky test using nitransform's resampling

oesteban:


jhlegarreta:
> I've gotten nitransforms out of the way

This is progress.

Thoughts:
- Why did we chose 0.25? It seems an arbitrary value, although I agree that we prefer to have low values as the upper error boundary.
- What does the HDF5 file contain? Are we able to reproduce its creation exactly?
- Does this happen if we use some other data stored in an HDF5 file?
- In the past, for some reason, when we've changed the `conftest.py` file, and even of seeds have not changed, some tests were affected, e.g. https://github.com/nipreps/nifreeze/pull/240#issue-3460743763. In this case it should not matter, because the `TrivialModel` instance returns the same map always, but maybe we've merged some PR modifying `conftest.py` and this is failing since? I still do not understand why this has happened in the past, but maybe we are facing the same issue here.
- Does this happen if we use any other iterator?
- Can you reproduce this locally?
  If yes,
  - Come back to the most recent stable commit where this was not happening and inspect the diff.
  If not,
  - Create a test repository to push the most recent stable commit where this was not happening and inspect the diff across them.
  - Not sure what we are missing to reproduce the GitHub environment exactly :-S.

oesteban:
I'm going to merge because although this doesn't fix the issue, it does reduce the problem (by getting rid of nitransforms).

PR #323, opened 2025-11-12
FIX: Refactor registration patching with monkeypatch

oesteban:
This mock potentially overloaded the _run_registration method for all tests run within the same thread.

### DIFF
--- a/test/test_estimator.py
+++ b/test/test_estimator.py
@@ -184,7 +184,11 @@ def test_estimator_iterator_index_match(
     class DummyXForm:
         matrix = np.eye(4)
 
-    nifreeze.estimator._run_registration = lambda *a, **k: DummyXForm()
+    monkeypatch.setattr(
+        nifreeze.estimator,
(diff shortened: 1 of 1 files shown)


jhlegarreta:
OK, I see. I contributed that code, so I take the blame for it. Thanks for all the investigation.

As for the remaining failure, since this error is also recent and we have not changed any version, maybe some other monkey patching that I have done in some other test results in this? Apologies if this ends up being the case :worried:.

Otherwise, wondering if the error comes from ANTs failing to converge and the underlying error being masked due to some limitation that makes such that `nipype` only informs about a later error in the chain (i.e. the transformation matrix not being written and, thus, not being found)?

`main` was passing consistently until commit adf71da, which corresponds to the merge commit of PR #299. PR #300 seems to be the first where the transformation matrix file not being present appears, https://github.com/nipreps/nifreeze/actions/runs/18957141877/job/54252874305. However, that PR does not seem to be the culprit to me.

PR #325, opened 2025-11-13
ENH: Adopt row-major gradient tables

oesteban:
## Summary
- switch the DWI data structure and serialization helpers to use row-major (N×C) gradient tables
- update diffusion models, shell-selection utilities, and unit tests to follow the new layout
- refresh usage documentation and the data-structures notebook to describe the row-major convention

## Testing
- pytest test/test_filtering.py::test_dwi_select_shells -q *(fails: ModuleNotFoundError: nifreeze._version)*

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_6915d241a8988330add559b8780a4cc8)

Fixes #316.

PR #327, opened 2025-11-17
REF: Refactor gradient data checks in DWI data class

jhlegarreta:
Refactor gradient data checks in DWI data class: PR https://github.com/nipreps/nifreeze/pull/325 introduced the row-major convention for gradient data in NiFreeze. However, gradient loading was not tested thoroughly. This resulted in some execution flows that would not guarantee a row-major internal convention or would crash under some circumstances.

This commit refactors the gradient data checks and adds thorough testing:
- Define all error or warning messages as global variables so that they can be checked exactly in tests.
- Ensure that gradients conform to the row-major convention when instantiating the DWI class directly. This allows to separate the gradient reformatting from the dimensionality check with the DWI volume sequence. This simplifies the flow, as the gradient reformatting to the row-major convention does not depend on the number of volumes in the DWI sequence. Also, this makes the flow more consistent with the refactored checks of the NIfTI file-based loading utility function (`from_nii`).
- Ensure that the gradients conform to the row-major convention immediately after loading the gradients file in the NIfTI file-based loading utility function (`from_nii`). As opposed to the previous implementation, this allows to load the gradients from a file where data follows either column-major or row-major convention. e.g. In the previous implementation the `if grad.shape[1] < 2:` was making an assumption about the layout and/or one that was wrong because we require 4 columns (direction (x,y,z) + b-value) or rows (if in column-major). The new implementation simplifies the execution flow.

### DIFF
--- a/src/nifreeze/data/base.py
+++ b/src/nifreeze/data/base.py
@@ -479,51 +479,82 @@ class BaseDataset(Generic[Unpack[Ts]]):
 
         """
 
-        if filename is None and write_hmxfms:
-            warn("write_hmxfms is set to True, but no filename was provided.", stacklevel=2)
-            write_hmxfms = False
+        return to_nifti(
+            self,
+            filename=filename,
+            write_hmxfms=write_hmxfms,
+            order=order,
+        )
 
-        if self.motion_affines is not None:  # resampling is needed
-            reference = ImageGrid(shape=self.dataobj.shape[:3], affine=self.affine)
-            resampled = np.empty_like(self.dataobj, dtype=self.dataobj.dtype)
-            xforms = LinearTransformsMapping(self.motion_affines, reference=reference)
 
-            # This loop could be replaced by nitransforms.resampling.apply() when
-            # it is fixed (bug should only affect datasets with less than 9 orientations)
-            for i, xform in enumerate(xforms):
-                frame = self[i]
-                datamoving = nb.Nifti1Image(frame[0], self.affine, self.datahdr)
-                # resample at index
-                resampled[..., i] = np.asanyarray(
-                    apply(xform, datamoving, order=order).dataobj,
-                    dtype=self.dataobj.dtype,
-                )
+def to_nifti(
+    dataset: BaseDataset,
+    filename: Path | str | None = None,
+    write_hmxfms: bool = False,
+    order: int = 3,
+) -> nb.Nifti1Image:
+    """
+    Write a NIfTI file to disk.
 
-                if filename is not None and write_hmxfms:
-                    # Prepare filename and write out
-                    out_root = Path(filename).absolute()
-                    out_root = out_root.parent / out_root.name.replace(
-                        "".join(out_root.suffixes), ""
-                    )
-                    xform.to_filename(out_root.with_suffix(".x5"))
-        else:
-            resampled = self.dataobj
+    Volumes are resampled to the reference affine if motion affines have
+    been set, otherwise the original data are written.
 
-            if write_hmxfms:
-                warn(
-                    "write_hmxfms is set to True, but no motion affines were found. Skipping.",
-                    stacklevel=2,
-                )
+    Parameters
+    ----------
+    filename : :obj:`os.pathlike`, optional
+        The output NIfTI file path.
+    write_hmxfms : :obj:`bool`, optional
+        If ``True``, the head motion affines will be written out to filesystem
+        with BIDS' X5 format.
+    order : :obj:`int`, optional
+        The interpolation order to use when resampling the data.
+        Defaults to 3 (cubic interpolation).
 
-        if self.datahdr is None:
-            hdr = nb.Nifti1Header()
-            hdr.set_xyzt_units("mm")
-            hdr.set_data_dtype(self.dataobj.dtype)
-        else:
-            hdr = self.datahdr.copy()
+    """
 
-        nii = nb.Nifti1Image(resampled, self.affine, hdr)
-        if filename is not None:
-            nii.to_filename(filename)
+    if filename is None and write_hmxfms:
+        warn("write_hmxfms is set to True, but no filename was provided.", stacklevel=2)
+        write_hmxfms = False
 
-        return nii
+    if dataset.motion_affines is not None:  # resampling is needed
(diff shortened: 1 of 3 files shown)


jhlegarreta:
A few notes:
- https://github.com/nipreps/nifreeze/blob/c729280d064a15ddcfc23e9666c12b3a83ba0b19/src/nifreeze/data/dmri.py#L90 should not be necessary/`dataobj` would not be allowed to be `None` after PR #305 gets merged.
- This line is never hit
https://github.com/nipreps/nifreeze/blob/c729280d064a15ddcfc23e9666c12b3a83ba0b19/src/nifreeze/data/dmri.py#L392
I'd assume that this could be `[x1 y1 z1 x2 y2 z2 ... xN yN zN]` (and its `.T` counterpart) gradient file. I'd say that reshaping that to an `Nx3` that array was not implemented in the previous implementation, and I have not implemented it. Would leave that for a separate PR.
- `_normalize_gradients`: it looks a misleading name now, as it does not normalize gradients. By the way, we should normalize gradients as well. That function looks the appropriate place, but as it performs other reformatting operations, we should come up with a more descriptive/comprehensive name. I've written the test for when we will tackle the gradient normalization: https://gist.github.com/jhlegarreta/9ffab029e96459846eaea27936e47c33
- The use of the b0s surfaces here again: when instantiating a DWI object directly (e.g. without the `from_nii`: have not checked), some user may very well do
 ```
    dwi = DWI(dataobj=dwi_dataobj, bzero=b0_dataobj, gradients=gradients)
    assert np.allclose(gradients, dwi.gradients)    
```

where `gradients` contain the b0s. The above would not fail because the `DWI` instantiation case does not filter the null gradients (does not use/accept `b0_thres`, as opposed to the `from_nii` counterpart:
```
    dwi = from_nii(dwi_fname, gradients_file=grads_fname, b0_thres=b0_thres)
    assert np.allclose(gradients[gradmask], dwi.gradients)
```

`dwi.gradients` would be e.g. `12x4` in the first case, which would not agree with what we would get when using the `from_nii` function, i.e. `10x4` (bzeros are removed).

The above is an inconsistency that needs to be addressed. I know that the instantiation case should not provide the b0s, but we cannot guarantee that it will not happen, even if we document the expected behavior very well. Here is a testing function for that: https://gist.github.com/jhlegarreta/1a3fbbfab362a2f2dbd28a8c53c54c08

Although `_normalize_gradients` looks an appropriate candidate to remove the b0s, in `from_nii` it is called
https://github.com/nipreps/nifreeze/pull/327/files#diff-774e30bff24e702cda5516cbc6bbbcae1af21b710e2b0ad22cc15741fbed9afeR430
 once the filtering has taken place
https://github.com/nipreps/nifreeze/pull/327/files#diff-774e30bff24e702cda5516cbc6bbbcae1af21b710e2b0ad22cc15741fbed9afeR439

which would mean that we may attempt to do the same job twice. Whatever way we take to solve this, that should be addressed in a separate PR.

**Comment by jhlegarreta** [comment]
Re: https://github.com/nipreps/nifreeze/pull/332#issuecomment-3560815841 So, the relevant changes from PR #305 are brought here. PR #305 can be safely closed.

File: src/nifreeze/data/dmri.py [hunk 1]
```diff
@@ -37,76 +37,118 @@
...
-                n_volumes = None
-
-        if n_volumes is not None and gradients.shape[0] != n_volumes:
-            if gradients.shape[1] == n_volumes:
-                gradients = gradients.T
-            else:
-                raise ValueError(
-                    "Gradient table shape does not match the number of diffusion volumes: "
-                    f"expected {n_volumes} rows, found {gradients.shape[0]}"
+            raise ValueError(GRADIENT_DATA_MISSING_ERROR)
```
oesteban: This is now addressed by the validator. Fixing.

src/nifreeze/data/dmri.py
```diff
@@ -37,76 +37,118 @@
...
+        ...
+        ValueError: Gradient table contains NaN or infinite values.
+        >>> validate_gradients(None, None, [[np.nan, 0.0, 0.0, 1000]])
+        Traceback (most recent call last):
+        ...
+        ValueError: Gradient table contains NaN or infinite values.
+
+    """
+    if np.shape(value)[1] != 4:
+        raise ValueError(GRADIENT_EXPECTED_COLUMNS_ERROR_MSG)
```
oesteban: The formatter will never allow this condition to occur. Adding a doctest to exercise this protection.

src/nifreeze/data/dmriutils.py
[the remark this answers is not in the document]
oesteban: Okay, I'm going all the way with the new submodule, I hate the `dmriutils` solution.

src/nifreeze/data/dmri/utils.py
```diff
@@ -0,0 +1,317 @@
...
+        raise TypeError(GRADIENT_OBJECT_ERROR_MSG) from exc
+
+    if formatted.ndim != 2:
+        raise ValueError(GRADIENT_NDIM_ERROR_MSG)
+
+    # If the numeric values are all integers, preserve integer dtype
+    if np.allclose(formatted, rounded := np.rint(formatted)):
+        formatted = rounded.astype(
+            formatted.dtype if np.issubdtype(formatted.dtype, np.integer) else np.int16
+        )
```
jhlegarreta: Although it was myself who introduced a version of this to make doctests pass,
https://github.com/nipreps/nifreeze/pull/327/commits/064413eb3fb9e8deaa8da7190265f2ee502ffc41#diff-774e30bff24e702cda5516cbc6bbbcae1af21b710e2b0ad22cc15741fbed9afeR178
I later realized that since we will be normalizing gradients, gradients will anyway be returned as floats. 
Also, it did not make much sense to do this just to make doctests pass. In https://github.com/nipreps/nifreeze/pull/327/commits/7933b75e4f3c7623572b1fed61e2c9740d59e3b0#diff-774e30bff24e702cda5516cbc6bbbcae1af21b710e2b0ad22cc15741fbed9afeR119, I ended up by setting the NumPy formatter in an explicit way to make test pass and removed the block. Hope this makes sense.

PR #331, opened 2025-11-19
FIX: Use the number of levels for overriding settings control branch

jhlegarreta:
Use the number of levels instead of the set of levels for overriding settings control branch: makes single-level settings be nested into lists.


jhlegarreta:
@oesteban I guess this is a bug fix. Would be grateful if you could come up with a test to avoid introducing regressions for this.

oesteban (approved):

jhlegarreta:
Merging. Will open an issue to keep track of the missing test.

PR #336, opened 2025-11-21
ENH: Validate PET data objects' attributes at instantiation

jhlegarreta:
Validate PET data objects' attributes at instantiation: ensures that the attributes are present and match the expected dimensionalities.

**PET class attributes**
Refactor the PET attributes so that only `midframe` and `total_duration` are required and accepted by the constructor. These are the only parameters that are required by the current PET model.

Remove `uptake` from the constructor: the PET data class does not need to know the uptake values held across its frames; it is rather the estimator that needs to know about its values so that the iterator can pick the frames following the appropriate sorting.

Validate and format attributes so to avoid missing or inconsistent data. Specifically, require the midframe data to have the same length as the number of frames in the data object, and disallow the last midframe value being larger than the total duration.

Make the `_compute_uptake_statistic` public so that users can call it.

**`from_nii`** function:
Refactor the `from_nii` function to accept filenames instead of a mix of filenames (e.g. the PET image sequence and brainmask) and temporal attribute arrays. Honors the name of the function, increases consistency with the dMRI counterpart and allows to offer a uniform API.

The only required temporal parameter required by BIDS is the frame time (`FrameTimesStart`). Thus, the temporal attribute JSON (sidecar) file is required to contain that key. The values required to model a PET dataset for the purposes of NiFreeze, namely the midframe and total duration values, are computed from the frame time. It is assumed that the frame duration spans entirely the time elapsed between two consecutive time frame values.

Refactor and rename the `_compute_frame_duration` function so that it computes and returns the required parameters to instantiate a PET data object. The computation of the relevant temporal values is, thus, done at this place only.

Use the `get_data` utils function in `from_nii` to handle automatically the data type when loading the PET data.

**`to_nifti`** function
Preserve the base class `to_nifti` method to serialize the `PET` dataset to NIfTI data. The `PET` dataset class does not need to write its temporal attributes, as they do not change along the prediction process, and they can be computed from the primary JSON file where `FrameTimesStart` data dwell. NiFreeze will still allow writing a BIDS-compatible derivative dataset by only writing the motion-corrected PET frames, and users would read the `FrameTimesStart` data from the primary JSON file.

**`PET.load`** class method:
Remove the `PET.load` class method and rely on the `data.__init__.load` function:
- If an HDF5 filename is provided, it is assumed that it hosts all necessary information, and the data module `load` function should take of loading all data.
- If the provided arguments are NIfTI files plus other data files, the function will call the `pet.PET.from_nii` function.

Change the `kwargs` arguments to be able to identify the relevant keyword arguments that are now present in the `from_nii` function.

Change accordingly the `PET.load(pet_file, json_file)` call in the PET notebook and the `test_pet_load` test function.

**Tests**:
Refactor the PET data creation fixture in `conftest.py` to accept the `frame_time` (as it is the only required arguments by BIDS and the one that allows computing the rest) and to return the necessary data.

Remove values that are no longer needed (i.e. `total_duration`).

Refactor the tests accordingly and increase consistency with the `dmri` data module testing helper functions. Reduces cognitive load and maintenance burden.

Add additional object instantiation equality checks: check that objects instantiated through reading NIfTI files equal objects instantiated directly.

Check the PET dataset attributes systematically in round trip tests by collecting all named attributes that need to be tested.

Modify accordingly the PET model and integration tests.

Take advantage of the patch set to make other opinionated choices:
- Prefer using the global `setup_random_pet_data` fixture over the local `random_dataset` fixture: it allows to control the parameters of the generated data and increases consistency with the practice adopted across the dMRI dataset tests. Remove the `random_dataset` fixture.
- Prefer using `assert np.allclose` over `np.testing.assert_array_equal` for the sake of consistency

**Dependencies**
Require `attrs>24.1.0` so that `attrs.Converter` can be used.
Documentation:
https://www.attrs.org/en/25.4.0/api.html#converters

### DIFF
--- a/docs/notebooks/pet_motion_estimation.ipynb
+++ b/docs/notebooks/pet_motion_estimation.ipynb
@@ -10,7 +10,7 @@
     "from os import getenv\n",
     "from pathlib import Path\n",
     "\n",
-    "from nifreeze.data.pet import PET\n",
+    "from nifreeze.data.pet import from_nii\n",
     "\n",
     "# Install test data from gin.g-node.org:\n",
     "#   $ datalad install -g https://gin.g-node.org/nipreps-data/tests-nifreeze.git\n",
@@ -29,7 +29,7 @@
     "    DATA_PATH / \"pet_data\" / \"sub-02\" / \"ses-baseline\" / \"pet\" / \"sub-02_ses-baseline_pet.json\"\n",
     ")\n",
     "\n",
-    "pet_dataset = PET.load(pet_file, json_file)"
+    "pet_dataset = from_nii(pet_file, temporal_file=json_file)"
    ]
   },
   {
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ classifiers = [
 license = "Apache-2.0"
 requires-python = ">=3.10"
 dependencies = [
-  "attrs>=20.1.0",
+  "attrs>=24.1.0",
   "dipy>=1.5.0",
   "joblib",
   "nipype>=1.5.1,<2.0",
--- a/src/nifreeze/data/__init__.py
+++ b/src/nifreeze/data/__init__.py
@@ -76,7 +76,7 @@ def load(
         from nifreeze.data.dmri import from_nii as dmri_from_nii
 
         return dmri_from_nii(filename, brainmask_file=brainmask_file, **kwargs)
-    elif {"frame_time", "frame_duration"} & set(kwargs):
+    elif {"temporal_file"} & set(kwargs):
         from nifreeze.data.pet import from_nii as pet_from_nii
 
         return pet_from_nii(filename, brainmask_file=brainmask_file, **kwargs)
--- a/src/nifreeze/data/pet.py
+++ b/src/nifreeze/data/pet.py
@@ -28,7 +28,7 @@ import json
 from collections import namedtuple
 from collections.abc import Callable
 from pathlib import Path
-from typing import Any
+from typing import Any, Tuple
 
 import attrs
 import h5py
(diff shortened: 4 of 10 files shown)


jhlegarreta:
Depends on PR #335.

mnoergaard:
> @mnoergaard While working on this I've realized that as things stand now on `main`, there is a risk that two instances of a PET object contain different data if instantiated directly (i.e. `PET()`) or from a NIfTI file (i.e. `from_nii`). So, I would like to have your confirm on the following:
> 
> - To build a valid PET instance, the only required piece of information are the data sequence and `frame_time`, as `uptake` can be computed from the former, and `midframe` and `total_duration` can all be computed from the latter.
> 
> 
> 
> I think that answer will allow me to refactor this properly and avoid inconsistencies. Thanks.

@jhlegarreta - that is correct! Thanks.

jhlegarreta:
Re https://github.com/nipreps/nifreeze/pull/336#issuecomment-3567033504 as of commit 37ed54c7a7dea53cd6fd4441cdd017d6af679505 and the `init=False`, I dug into this a little more:
- I see that `to_filename` dumps the PET class instance entirely to the HDF5 files, including its private attributes (i.e. those that can be computed from the data: `midframe`, `total_duration`, etc.) Thus, when a `PET` object is tried to be read and instantiated from the HDF5 contents, as it is given all these data, including the private attributes, it fails.
Due to the way uptake is computed (using a callable that is not stored anywhere), and since it does not make sense to instantiate a PET object to have the private attributes computed to immediately after set them to other values, best would be to allow all of them to be present in the constructor, falling back to the default way of computing them if not present.
- I see that that `load` class method is taking a JSON file with the frame duration and frame time start data. The frame time start looks like it should be an attribute of the class that can be given at instantiation, and be defaulted to
  ```
  frame_time_arr = np.array(self.frame_time, dtype=np.float32)
  frame_time_arr -= frame_time_arr[0]
  ```
  if not given.

Also, it should be possible to host and read midframe and total duration data from a JSON file (or any other file), much like it is done with the frame duration data, falling back to the default way of computing them if not present.

Refactoring things this way (i.e. allowing to provide all attributes to the `PET` class, and the `from_nii` function, and falling back to defaults if not given) would probably allow us to solve all failing tests that persist.

WDTY @mnoergaard?

Sorry for so many questions. Hopefully we will converge and the implementation across the multiple ways to read/write data will be consistent and robust after this.

jhlegarreta:
Marking this as a draft: commits 747b0028876772c9228c68cb8c6480e022897044 and 387dfe63c2ae1e59f914c8f93beb4277bb191a56 show that the difficulties to have a robust PET data instantiation (i.e. not requiring more than what we need while trying to disallow means to set parameter values that can be wrong or inconsistent across time arrays), having a consistent API across modalities and being BIDS-compatible when serializing/reading the data required by the class).

I think the PET estimator (PR #203) needs to be fixed so that this is not necessary:
https://github.com/nipreps/nifreeze/blob/f892e52ac24f09c49c458d81f3bdca86c78b2b7f/src/nifreeze/estimator.py#L256-L261

We should not need to do that instantiation, which conditions in important ways the PET data class refactoring.

Commit https://github.com/nipreps/nifreeze/pull/336/commits/dd3738a0b2cb548c0de29ab70562d7ec8e5868b4 shows why `frame_time` would be required to serialize a BIDS-valid dataset, and the need to reconstruct it from `midframe` and `total_duration`.

I have explored a lot of other avenues to try to have this sorted and none is robust. PRs #203 and #204 need to be worked on and merged before more time is invested in this PR.

PR #382, opened 2025-12-24
MNT: Continue running tests until completion

jhlegarreta:
Continue running tests until all have run: remove the `x` flag from the `pytest` command to avoid stopping test runs immediately after the first failure or error (fail-fast behavior).

Running all tests until completion allows to identify all side-effects of a given change at once, without waiting for multiple CI cycles, saving developer time.


jhlegarreta:
Type checking errors and build warnings are unrelated. Merging.

PR #423, opened 2026-02-08
ENH: Implement start/end index for the estimator

jhlegarreta:
Implement start/end index for the estimator.


jhlegarreta:
@oesteban The last commit makes the modality-specific keyword arguments take precedence over `size`, as agreed this morning. It also ships some other improvements (apologies for not separating these into different commits).

jhlegarreta:
The regression in `PETBenchmark.time_getitem` (and I believe I have seen it on `DWIBenchmark.time_getitem` once) seems real as it is consistent; around 1.3x slower. At this point, I prefer to have this merged as is, as it has been there for almost 3 months and has been a recurrent theme lately; if you feel like the performance should be improved, please take over @mnoergaard @oesteban.

mnoergaard (approved):

PR #440, opened 2026-04-11
REF: Avoid divide by zero warnings when denominator is zero

jhlegarreta:
Avoid divide by zero warnings when denominator is zero: check that values are nonzero before dividing; else, set all values to zero.

Add the corresponding tests.

Fixes:
```
test/test_integration.py: 15 warnings
    /home/runner/work/nifreeze/nifreeze/src/nifreeze/data/filtering.py:107:
    RuntimeWarning: invalid value encountered in divide
          data /= data.max()
```

raised for example at:
https://github.com/nipreps/nifreeze/actions/runs/24288044855/job/70920601980?pr=313

### DIFF
--- a/src/nifreeze/data/filtering.py
+++ b/src/nifreeze/data/filtering.py
@@ -104,7 +104,12 @@ def advanced_clip(
     # Clip and scale data
     np.clip(data, a_min=a_min, a_max=a_max, out=data)
     data -= data.min()
-    data /= data.max()
+    rng = float(data.max())  # rng = max-min
+    if rng != 0.0:
+        data /= rng
+    else:
+        # All values are equal after clipping: normalized result should be zeros
+        data.fill(0.0)
 
     if invert:
         np.subtract(1.0, data, out=data)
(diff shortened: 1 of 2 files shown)


src/nifreeze/data/filtering.py
```diff
@@ -104,7 +104,12 @@ def advanced_clip(
     # Clip and scale data
     np.clip(data, a_min=a_min, a_max=a_max, out=data)
     data -= data.min()
-    data /= data.max()
+    rng = float(data.max())  # rng = max-min
```
arokem:
```suggestion
    _range = float(data.max())  # rng = max-min
```

src/nifreeze/data/filtering.py
```diff
@@ -104,7 +104,12 @@ def advanced_clip(
...
+    if rng != 0.0:
+        data /= rng
```
arokem:
```suggestion
        data /= _range
```

arokem (commented):
Good idea, but a couple of comments on implementation

mnoergaard (approved):

PR #443, opened 2026-06-09
FIX: Parallelize DKI Fit/Predict for Usable Multi-Shell Motion Estimation

### LINKED ISSUE #442: DKI head-motion estimation is impractically slow (~8.6 vol/h, ~32 h/series)
## Summary

Used as the head-motion/eddy estimator for dMRI, the **DKI** model is effectively unusable on real multi-shell data, while **DTI** and **trivial** on the *same data* run fine. The DKI code path serially re-fits a full-brain kurtosis model on every leave-one-volume-out (LOVO) iteration, with BLAS threads left uncontrolled.

## Symptom / throughput

- Data: 104×146×87×**280**, **274 weighted** volumes after b0 handling, **4 shells** (b=700/1000/2000/3000) + 6 b0, brain mask ≈ 388,196 voxels. NiFreeze `26.0.0.dev894`.
- **27 of 274** per-volume `ants-*GenericAffine.mat` transforms written in **3h09m** ⇒ ~**8.6 vol/h** ⇒ ~**32 h**/series; the process appears "hung."
- Not a deadlock/OOM/thrash: main process `R (running)`, ~**489% CPU across 72 threads**, RSS ~2 GB (31 GB host), ~21 non-voluntary context switches / 2 s → genuinely compute-bound, with BLAS spread across all cores.

## Reproduce

```python
from nifreeze.data.dmri import from_nii
from nifreeze.estimator import Estimator
dwi = from_nii("degibbs.nii.gz", bvec_file="checked.bvec", bval_file="checked.bval",
               brainmask_file="b0ref_brain_mask.nii.gz", b0_file="b0ref_undistorted.nii.gz")
Estimator("dki", strategy="random").run(dwi, n_jobs=8, seed=42)   # <-- crawls
# Estimator("dti", ...) on the same dwi is fast.
```

(Benign-looking `UserWarning` on load: *"DWI data contains b0 volumes, but the 'bzero' attribute was set. DWI b0 volumes will be discarded…"* — because `b0_file=` was passed alongside in-data b0s.)

## Root cause (confirmed by reading the code)

All in `src/nifreeze/model/dmri.py`:

1. **DKI hard-forced onto the serial full-brain path** — line ~307, `if n_jobs == 1 or is_dki:` routes DKI to a single serial `_exec_fit` over the whole brain, bypassing the `np.array_split`-over-voxels + `joblib.Parallel` chunking that DTI uses (lines ~316–330). Voxel-wise chunking is embarrassingly parallel and **exact** for DKI too, so the special-case looks unjustified for the joblib path. Introduced in `d513d057` (#157), with the comment *"DKI model does not allow parallelization as implemented here"* — but that justification traces to #142's *different* approach (dipy's internal `multi_voxel_fit` engine), not array-split chunking.
2. **Dead code** — the `elif is_dki:` branch (lines ~311–314, `model.multi_fit(...)`) is unreachable: the preceding `if` already consumes `is_dki`.
3. **Per-LOVO full refit** — `Estimator.run` calls `fit_predict(i)` for all 274 volumes; `_fit`'s `_locked_fit` short-circuit (line ~259) only triggers when `index is None`, which the default loop never sends. A fit-once-reuse mode exists (`Estimator(single_fit=True)`) but is off by default and not wired for DKI. DKI's per-fit cost (22-parameter kurtosis WLS) dwarfs DTI's, so refitting ×274 is the dominant real cost.
4. **BLAS oversubscription** — `omp_nthreads` is popped and discarded (line ~343); `threadpoolctl` is used nowhere and is not even a dependency. A single full-brain DKI fit lets BLAS spin all cores → the observed 489% CPU / 72 threads. Compounds (1)+(3).

## Questions

1. **Why** is DKI forced serial — numerical stability, dipy thread-safety, or a historical artifact of #142's abandoned `multi_voxel_fit` route?
2. Is the per-LOVO refit avoidable via `single_fit` (fit-once, predict-each)? Is the LOVO-vs-single approximation acceptable for HMC with 274 volumes?
3. Can DKI safely use DTI's array-split joblib chunking (voxel-wise = exact)? If not, what minimal change parallelizes it?
4. Is BLAS oversubscription (uncontrolled OMP/OPENBLAS × joblib) compounding it — would honoring `omp_nthreads` via `threadpool_limits` help?
5. Is DKI even the right HMC predictor here, or should multi-shell default to a faster model (GP/DTI), gating DKI behind explicit opt-in + a warning?

## Proposed fix options (menu)

- **A. Enable voxel-chunked joblib for DKI** — drop the `is_dki` serial special-case and remove the dead `elif`. Likely exact and the biggest single win.
- **B. Wire `single_fit` for DKI** (fit-once, predict-each), as default or opt-in for kurtosis — removes the ×274 refit.
- **C. Honor `omp_nthreads` via `threadpoolctl.threadpool_limits`** around the fit to stop BLAS oversubscription (adds a `threadpoolctl` dependency).
- **D. Gate DKI behind opt-in** for multi-shell HMC with a perf warning; default to GP/DTI.

Cross-refs: #158, #157, #142, #434.

oesteban:
## Summary

Used as the head-motion/eddy estimator for dMRI, the **DKI** model was effectively unusable on real multi-shell data (~8.6 vol/h ⇒ ~32 h/series, appearing hung), while **DTI** on the same data runs in minutes. This PR makes DKI parallelize correctly, recovering a ~5.8× speedup (more with more workers) with **numerically identical** results.

Resolves: #442

## Root cause

DKI was hard-routed onto the serial full-brain fit path (`if n_jobs == 1 or is_dki:`), so per-volume cost grew linearly with voxel count while DTI's `np.array_split` + `joblib` path stayed flat (measured 6×→40×→119× slower as voxels grew 2k→20k→80k). The serial special-case existed for a real reason: **DKI's `MultiVoxelFit` cannot be pickled** across loky's process boundary, so naively reusing DTI's fit/predict-split path crashes (`BrokenProcessPool` / `RecursionError`). BLAS thread-capping was ruled out as a latency factor (no effect — the fit is GIL-bound per-voxel Python work). Full diagnosis with timings in #442.

## Fix

Add an exact, in-worker parallel path for models whose fitted object cannot cross a process boundary:

- New `_exec_fit_predict(...)` worker: builds the model, fits its voxel chunk, predicts the held-out gradient, and returns **only the predicted ndarray** — no model instance is ever serialized.
- New `BaseDWIModel._fit_predict_chunked(index, n_jobs)`: splits voxels (and the aligned `S0`) across workers, runs the worker per chunk, and re-assembles the prediction. Because voxel-wise fitting is independent, the result is **bit-identical** to the serial path.
- `fit_predict` routes to this path when `not self._picklable_fit and n_jobs > 1`. `DKIModel` sets `_picklable_fit = False`; all other models keep the existing behavior.
- Removed the unreachable `elif is_dki:` (dead `model.multi_fit` branch) and factored the shared data prep into `_lovo_data`.

The DTI fit/predict-split path is untouched. DKI at `n_jobs == 1`, direct `_fit` calls, and `single_fit` are unchanged.

## Benchmark (synthetic multi-shell, b0 + 3 shells, n_jobs=8)

| voxels | DKI serial (before) | DKI parallel (after) | speedup |
|-------:|--------------------:|---------------------:|--------:|
| 20,000 | 19.3 s/vol | 3.3 s/vol | 5.8× |

Per-volume output matches the serial path exactly (`max|diff| = 0`).

## Tests

- New `test_dki_parallel_matches_serial` (`index ∈ {4,9} × use_mask ∈ {False,True}`) asserts `fit_predict(index, n_jobs=4)` (chunked) equals `fit_predict(index, n_jobs=1)` (serial).
- Full `test_model_dmri.py` DKI/DTI suite green (132 passed).

## Notes

- `single_fit` (fit-once, predict-each) could cut runtime further but is approximate and currently buggy for `n_jobs > 1`; left for a follow-up.
- `CHANGES.rst` intentionally untouched (release-generated).

### DIFF
--- a/src/nifreeze/model/dmri.py
+++ b/src/nifreeze/model/dmri.py
@@ -55,6 +55,23 @@ def _exec_predict(model, chunk=None, **kwargs):
     return np.squeeze(model.predict(**kwargs)), chunk
 
 
+def _exec_fit_predict(model_class, bvals, bvecs, data, gradient, chunk=None, S0=None, **kwargs):
+    """Fit a model and predict a held-out gradient within a single worker.
+
+    This is used for models whose fitted object cannot be pickled across a
+    process boundary (e.g. DKI's :obj:`~dipy.reconst.multi_voxel.MultiVoxelFit`).
+    The fit happens inside the worker and only the predicted array is returned,
+    so no model instance is ever serialized. See gh-442.
+    """
+    module_name, class_name = model_class.rsplit(".", 1)
+    fit_gtab = gradient_table_from_bvals_bvecs(bvals, bvecs)
+    fitted = getattr(import_module(module_name), class_name)(fit_gtab, **kwargs).fit(data)
+    pred_gtab = gradient_table_from_bvals_bvecs(
+        gradient[np.newaxis, -1], gradient[np.newaxis, :-1]
+    )
+    return np.squeeze(fitted.predict(pred_gtab, S0=S0)), chunk
+
+
 def _compute_data_mask(
     shape: tuple,
     brainmask: np.ndarray | None = None,
@@ -198,6 +215,13 @@ class BaseDWIModel(BaseModel):
         "_models": "List with one or more (if parallel execution) model instances",
     }
 
+    _picklable_fit: bool = True
+    """Whether the fitted object can be pickled (sent across a process boundary).
+
+    Models whose fit cannot be pickled (e.g. DKI) use the in-worker
+    fit-and-predict path (:meth:`_fit_predict_chunked`) for parallelization.
+    """
+
     def __init__(self, dataset: DWI, max_b: float | int | None = None, **kwargs):
         """Initialization.
 
@@ -251,21 +275,26 @@ class BaseDWIModel(BaseModel):
 
         super().__init__(dataset, **kwargs)
 
-    def _fit(self, index: int | None = None, n_jobs: int | None = None, **kwargs) -> int:
-        """Fit the model chunk-by-chunk asynchronously"""
+    def _lovo_data(self, index: int | None = None):
+        """Assemble masked data and gradient table for a leave-one-volume-out fit.
 
-        n_jobs = n_jobs or 1
+        Parameters
+        ----------
+        index : :obj:`int` or ``None``
+            Volume index left out of the fit (``None`` keeps every volume).
 
-        if self._locked_fit is not None:
-            return n_jobs
+        Returns
+        -------
+        data : :obj:`~numpy.ndarray`
+            The masked (or flattened) signal, shaped ``(n_voxels, n_volumes)``.
+        gtab : :obj:`~dipy.core.gradients.GradientTable`
+            The matching gradient table (with b0 appended for DKI).
 
+        """
         brainmask = self._dataset.brainmask
         idxmask = np.ones(len(self._dataset), dtype=bool)
-
         if index is not None:
             idxmask[index] = False
-        else:
-            self._locked_fit = True
 
         data, _, gtab = self._dataset[idxmask]
         # Select voxels within mask or just unravel 3D if no mask
@@ -289,6 +318,22 @@ class BaseDWIModel(BaseModel):
                 )
             data, gtab = _append_bzero(data, gtab, bzero=bzero)
 
+        return data, gtab
+
+    def _fit(self, index: int | None = None, n_jobs: int | None = None, **kwargs) -> int:
+        """Fit the model chunk-by-chunk asynchronously"""
+
+        n_jobs = n_jobs or 1
+
+        if self._locked_fit is not None:
+            return n_jobs
+
+        if index is None:
+            self._locked_fit = True
+
+        data, gtab = self._lovo_data(index)
+
+        model_str = getattr(self, "_model_class", "")
         if model_str:
             module_name, class_name = model_str.rsplit(".", 1)
             model = getattr(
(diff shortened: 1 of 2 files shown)


oesteban:
Re: @arokem's suggestion to use DIPY's `multi_voxel` decorator (Ray/engine) — I tested it against the pinned DIPY (1.10.0) and it isn't usable on our current floor: `DKIModel.fit()` doesn't accept `engine`, and the reachable `multi_fit(engine=...)` path hits a decorator kwargs-leak bug (`engine` forwarded into `ls_fit_dki()`); `ray` also isn't a dependency. Details in nipreps/nifreeze#442 (comment). This PR's in-worker chunking is therefore the interim that works on `dipy>=1.5`; a code comment marks the DIPY-native path as the follow-up once a release supports it.

src/nifreeze/model/dmri.py
```diff
@@ -55,6 +55,23 @@ def _exec_predict(model, chunk=None, **kwargs):
...
 
+def _exec_fit_predict(model_class, bvals, bvecs, data, gradient, chunk=None, S0=None, **kwargs):
+    """Fit a model and predict a held-out gradient within a single worker.
+
+    This is used for models whose fitted object cannot be pickled across a
+    process boundary (e.g. DKI's :obj:`~dipy.reconst.multi_voxel.MultiVoxelFit`).
+    The fit happens inside the worker and only the predicted array is returned,
+    so no model instance is ever serialized. See gh-442.
+    """
+    module_name, class_name = model_class.rsplit(".", 1)
```
arokem, line 66: Maybe 
```suggestion
    module_name, class_name = model_class.rsplit(".", -1)
```

in case of `dipy.reconst.dti` and similar?

src/nifreeze/model/dmri.py
```diff
@@ -198,6 +215,13 @@ class BaseDWIModel(BaseModel):
         "_models": "List with one or more (if parallel execution) model instances",
     }
 
+    _picklable_fit: bool = True
+    """Whether the fitted object can be pickled (sent across a process boundary).
+
+    Models whose fit cannot be pickled (e.g. DKI) use the in-worker
```
arokem:
```suggestion
    Models whose fit cannot be pickled (e.g., DKI on DIPY 1.5) use the in-worker
```

arokem (approved):

src/nifreeze/model/dmri.py
oesteban: Since it's right split, should be 1, right?

jhlegarreta:
I guess having a DKI benchmarking test would be helpful here to see the gains:
https://github.com/nipreps/nifreeze/pull/446

I confess that the benchmark has been written in a haste, so it will require more work to demonstrate the issue.

arokem:
This can be closed, because fixes related to #442 and this PR have been addressed upstream in https://github.com/dipy/dipy/pull/4054

PR #448, opened 2026-07-08
MNT: Prefer matplotlb version bounding over compatibility specifier

jhlegarreta:
Prefer matplotlb version bounding over compatibility specifier introduced in 0c1d328b70f42b6afe7404a2d762dff3119aaa3c.

Use Python version-dependent specification to avoid trying to install incompatible matplotlib versions.

Fixes:
```
× No solution found when resolving dependencies:
╰─▶ Because the current Python version (3.10.20) does not satisfy
    Python>=3.11 and matplotlib==3.11.0 depends on Python>=3.11, we can
    conclude that matplotlib==3.11.0 cannot be used.
    And because only matplotlib<=3.11.0 is available and you require
    matplotlib>=3.11.0, we can conclude that your requirements are
    unsatisfiable.
```

raised, for example, in:
https://github.com/nipreps/nifreeze/actions/runs/27514840905/job/81321545725


jhlegarreta:
Errors are unrelated and will be addressed in a separate PR (#449). We will only know if this works once it is merged as Python 3.10 is not run per PR.

PR #450, opened 2026-07-09
FIX: Revise DKI model initialization to leverage DIPY's ``multi_voxel_fit``

oesteban:
After the patching of the decorator (dipy/dipy#4054), this PR adapts the model initialization to use DIPY's parallelization.

The PR includes tests to ensure the right fit functions inside DIPY's models are called.

### DIFF
--- a/src/nifreeze/model/dmri.py
+++ b/src/nifreeze/model/dmri.py
@@ -198,6 +198,12 @@ class BaseDWIModel(BaseModel):
         "_models": "List with one or more (if parallel execution) model instances",
     }
 
+    _multivoxel_engine: bool = False
+    """Whether the underlying model fits per-voxel through DIPY's
+    ``multi_voxel_fit`` and therefore supports DIPY-native parallelization
+    (``engine``/``n_jobs``). Models that do not (DTI, GQI, GP) are parallelized
+    by chunking the data across workers instead."""
+
     def __init__(self, dataset: DWI, max_b: float | int | None = None, **kwargs):
         """Initialization.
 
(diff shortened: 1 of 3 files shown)


src/nifreeze/model/dmri.py
```diff
@@ -291,6 +298,10 @@ def _fit(self, index: int | None = None, n_jobs: int | None = None, **kwargs) ->
 
         if model_str:
             module_name, class_name = model_str.rsplit(".", 1)
+
+            if self._multivoxel_engine and n_jobs > 1:
+                kwargs = kwargs | {"engine": "joblib", "n_jobs": n_jobs}
```
oesteban, line 306: A side effect of the double initialization of models - I think engine and n_jobs should be fit_kwargs?

oesteban: okay, I'm wrong - fit doesn't accept these arguments.

src/nifreeze/model/dmri.py
```diff
@@ -257,7 +266,7 @@ def _fit(self, index: int | None = None, n_jobs: int | None = None, **kwargs) ->
         n_jobs = n_jobs or 1
```
jhlegarreta, line 266: It is unclear the use of `_locked_fit`  after this change. Why does this return the number of stacked models whereas the other path returns the number of jobs https://github.com/nipreps/nifreeze/pull/450/changes#diff-b58005fd41fe310827c710a6540c00b56f75574cc68dba23d06d6ccb5a34c880L330?

oesteban: If the model has been parallelized and it's already locked, it means that you will have as many models as data chunks. I agree this line is not necessary anymore. If anything, this change is more reliable, TBH.

jhlegarreta: OK, then is not the number of stacked models, but the number of parallelized models. OK. Resolving.

jhlegarreta (approved):
Thanks for doing this, Óscar.

PR #452, opened 2026-07-09
RF: The adopton of multi_voxel_fit from DIPY is not immediate

oesteban:
DKI is the only method where fit is not decorated, and the parallelization parameters are passed at `__init__` time.

### DIFF
--- a/src/nifreeze/model/dmri.py
+++ b/src/nifreeze/model/dmri.py
@@ -272,9 +272,10 @@ class BaseDWIModel(BaseModel):
         # Select voxels within mask or just unravel 3D if no mask
         data = data[brainmask, ...] if brainmask is not None else data.reshape(-1, data.shape[-1])
 
-        # All BaseDWIModel subclasses expect a DIPY GradientTable.
-        model_str = getattr(self, "_model_class", "")
+        # Replace the gradient table with a DIPY object
         gtab = gradient_table_from_bvals_bvecs(gtab[:, -1], gtab[:, :-1])
+
+        model_str = getattr(self, "_model_class", "")
 
         # Append the b0 (if existing) to the gradients and the data for the
         # kurtosis model.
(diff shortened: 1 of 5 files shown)


test/test_model_dmri.py
```diff
@@ -1010,53 +1011,76 @@ def test_dki_model_predict(multi_shell_test_data, index, ignore_bzero, use_mask)
...
+            raise TypeError("boom")
+
+    dataset, _, _, _ = setup_multi_shell_fit_predict_data(
+        multi_shell_test_data, ignore_bzero=False, use_mask=False
+    )
+    monkeypatch.setattr("nifreeze.model.dki.DiffusionKurtosisModel", _RaisingDKIModel)
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message=MASK_ABSENCE_WARN_MSG, category=UserWarning)
+        with pytest.raises(TypeError, match="boom"):
```
jhlegarreta: Curious to know which package contains this "boom" message. It is rather vague as message.

src/nifreeze/model/dmri.py
```diff
@@ -299,37 +291,31 @@ def _fit(self, index: int | None = None, n_jobs: int | None = None, **kwargs) ->
...
-        if n_jobs == 1 or self._multivoxel_engine:
-            _modelfit, _ = _exec_fit(model, data, **fit_kwargs)
+        pargs = {} if n_jobs == 1 else {"engine": "joblib", "n_jobs": n_jobs}
+        try:
+            _modelfit, _ = _exec_fit(model, data, **pargs)
             self._models = [_modelfit]
             return 1
+        except TypeError:
+            if n_jobs == 1:
+                raise
```
jhlegarreta, line 308: Should this raise a more explicit message, or is it an underlying DIPY message that will be raised?

src/nifreeze/model/dmri.py
```diff
@@ -280,10 +272,10 @@ def _fit(self, index: int | None = None, n_jobs: int | None = None, **kwargs) ->
         # Select voxels within mask or just unravel 3D if no mask
         data = data[brainmask, ...] if brainmask is not None else data.reshape(-1, data.shape[-1])
 
-        # DIPY models (or one with a fully-compliant interface)
+        # Replace the gradient table with a DIPY object
+        gtab = gradient_table_from_bvals_bvecs(gtab[:, -1], gtab[:, :-1])
+
         model_str = getattr(self, "_model_class", "")
-        if "dipy" in model_str or "GeneralizedQSamplingModel" in model_str:
-            gtab = gradient_table_from_bvals_bvecs(gtab[:, -1], gtab[:, :-1])
```
jhlegarreta, line 286: Unsure why this change was related to the DKI mutli_voxel_fit thing. If unrelated, a separate PR/commit would be best IMO.

test/test_model_dmri.py
oesteban: It's the mock above https://github.com/nipreps/nifreeze/pull/452/changes#diff-392fe9846219097c6c9db5beb0e00b29b5357e7cd715585deb0d5b4a14264b47R1073

src/nifreeze/model/dmri.py
oesteban: This re-raises the TypeError as issued by DIPY.

src/nifreeze/model/dmri.py
oesteban: This patch would have added one more test here (if "dipy", if GQI, + if DKI), and the only actual model I can think of would be the GPR. Now, I looked at GPR, and it does look like we made it dipy-compatible too.

So, I thought it was safer to completely remove this branch, and when we come back to GPR, if it gives us a holler, we'll re-insert the object creation. WDYT?

test/test_model_dmri.py
oesteban: Let me know if "boom" is too annoying a message.

src/nifreeze/model/dmri.py
jhlegarreta: Fine, but it is still unrelated to the PR, and not needed for the DKI refactoring, so it makes the review more difficult, and becoming a source of confusion for the time we will need to revise why some changes were done. Keep it here now that it has been done.

test/test_model_dmri.py
jhlegarreta: OK, sorry, my bad for not realizing where it came from. I would discourage using "boom", but keep it in order to move forward. It can be changed in a separate PR.

jhlegarreta (approved):

PR #453, opened 2026-07-11
RF: Revise GQI, syncing with DIPY and adding Olivier's tests

oesteban:


src/nifreeze/model/gqi.py
```diff
@@ -107,68 +134,65 @@ def __init__(
...
             )
             @ self.model.kernel
         )
 
-        return (K @ self.data.T).T
+        return np.maximum((K @ self.data.T).T, 0)
 
 
 def gqi_kernel(gtab, param_lambda, sphere, method="standard"):
+    r"""Forward GQI kernel, shape ``(n_gradients, n_vertices)``.
```
oesteban:
```suggestion
    r"""Compute the forward GQI kernel, shape ``(n_gradients, n_vertices)``.
```

PR #457, opened 2026-07-13
FIX: Correct Log-Space GP Kernel Gradients and Make Noise Tunable via WhiteKernel

### LINKED ISSUE #456: GP hyperparameter optimization: fix log-space kernel gradients and make noise (σ²) tunable via WhiteKernel

oesteban:
## Summary

Fixes the GP kernel gradient computation and makes the measurement noise tunable, as described in the issue.

- **Log-space kernel gradients (bug).** `SphericalKriging`/`ExponentialKriging` returned gradients w.r.t. the hyperparameters directly, but scikit-learn's `Kernel.theta` is log-transformed and the optimizer expects `dK/d(log θ) = θ·dK/dθ`. Each channel is now scaled by its hyperparameter (for the scale `λ` the channel becomes `λ·C = K`). The single-shell GP runs the default L-BFGS optimizer with `eval_gradient=True`, so it was previously optimizing `beta_a`/`beta_l` with incorrect gradients. A finite-difference regression test now guards this (matches to ~1e-11).
- **Tunable noise via `WhiteKernel` (enhancement).** `GaussianProcessModel` folds the measurement noise into the covariance as `KernelType(...) + WhiteKernel(noise_level=sigma_sq)`, so it is optimized alongside the other hyperparameters instead of sitting in the fixed `alpha` nugget (whose kernel-gradient is zero). `sigma_sq` becomes the initial `noise_level`; `alpha` is kept as a small numerical jitter. This is more faithful to :cite:`andersson_non-parametric_2015`.
- Reworded the `DiffusionGPR` note about `alpha`, and added `scipy.*`/`sklearn.*` to the mypy `ignore_missing_imports` overrides (a new sklearn import is introduced).

**Behavioral change:** making the noise tunable shifts single-shell GP predictions (and hence motion estimates). Existing tests assert shapes/finiteness and still pass.

**Out of scope (later):** per-shell noise `σ_i²` (Eq. 16) for the multi-shell case is not a stock `WhiteKernel` — it needs a small custom kernel applying block-diagonal noise from the b-value column.

## Testing

- `pytest test/test_model_gpr.py test/test_model_dmri.py test/test_model.py` — all pass.
- Added: finite-difference gradient test for both Kriging kernels; a test asserting `GaussianProcessModel` adds a `WhiteKernel` whose `noise_level` is actually optimized during fit.
- `ruff` and `mypy` clean on the changed modules (one pre-existing `minimize` call-overload in `gpr.py` is untouched).

## Notes

- The gradient fix is a prerequisite for the multi-shell kernel work (#337 / #175), whose orientation sub-kernel reuses `SphericalKriging`; #175 should rebase on top of this.
- `main`'s "Unit and integration tests" workflow is already failing independently of this change.

Resolves: #456
X-Refs: #337, #175

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,6 +156,8 @@ module = [
   "joblib",
   "h5py",
   "ConfigSpace",
+  "scipy.*",
+  "sklearn.*",
 ]
 ignore_missing_imports = true
 
--- a/src/nifreeze/model/_dipy.py
+++ b/src/nifreeze/model/_dipy.py
@@ -36,9 +36,8 @@ from nifreeze.model.gpr import (
     SphericalKriging,
 )
 
-#: Small nugget kept on ``alpha`` for numerical stability once the measurement
-#: noise is modeled explicitly by a :obj:`~sklearn.gaussian_process.kernels.WhiteKernel`.
 GP_JITTER = 1e-10
+"""Small nugget kept on ``alpha`` for numerical stability."""
 
 
 def _cartesian_components(gtab: GradientTable | np.ndarray) -> np.ndarray:
@@ -152,9 +151,7 @@ class GaussianProcessModel(ReconstModel):
         self.sigma_sq = sigma_sq
 
         KernelType = SphericalKriging if kernel_model == "spherical" else ExponentialKriging
-        # Add the measurement noise as a WhiteKernel so it lives *inside* the
-        # covariance and is tuned by the analytical-gradient optimizer (a fixed
-        # ``alpha`` nugget has zero kernel-gradient and cannot be optimized).
+        # Add the :math:`\sigma^2` term of Andersson et al. (2015) as a WhiteKernel
         self.kernel = KernelType(
             beta_a=beta_a,
             beta_l=beta_l,
@@ -211,8 +208,6 @@ class GaussianProcessModel(ReconstModel):
             kernel=self.kernel,
             random_state=random_state,
             n_targets=y.shape[1],
-            # Noise is now carried by the WhiteKernel; keep ``alpha`` as a small
-            # numerical-stability jitter only.
             alpha=GP_JITTER,
         )
         self._modelfit = GPFit(
--- a/src/nifreeze/model/_dipy.py
+++ b/src/nifreeze/model/_dipy.py
@@ -28,12 +28,17 @@ import numpy as np
 from dipy.core.gradients import GradientTable
 from dipy.reconst.base import ReconstModel
 from sklearn.gaussian_process import GaussianProcessRegressor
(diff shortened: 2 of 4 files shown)


oesteban:
I'm fairly confident about this one. Merging.

PR #460, opened 2026-07-14
ENH: Prediction gallery of DWI schemes and models

### LINKED ISSUE #458: Generate a gallery of "predicted" volumes

It would be fantastic to add a "Gallery" to the documentation, showing for several real datasets (a simple, legacy DTI dataset, a single-shell HARDI dataset, a multi-shell dataset, and a DSI dataset) several "predicted" volumes, with each of the relevant models in both conditions (LOVO, holding out the predicted orientation and "single-fit").

For example, for DTI, show the average and DTI models; for single-shell HARDI, perhaps add GQI on top of those; then, for multi-shell, show DKI, GQI, and GPR; and finally, for DSI, show GQI.

It would be useful to track coverage when running these, so that we record what's been exercised.

Yes

_No response_

oesteban:
## Human summary

With Claude's help, I believe we are on the verge of a clear breakthrough. This PR generates a gallery of visualizations to assess model fitting. It fits 4 whole-brain images from 4 different OpenNeuro datasets. Each dataset has a different acquisition scheme: legacy DTI, single-shell HARDI (b=1000 though), multi-shell, and DSI. Each dataset is fitted with the models that are supposed to work for them.

Then, the gallery is added to the documentation and works as a canary in the coal mine:

- For example, the DTI predictions seem to have a DC problem. I've seen it a lot, and I've worked quite a bit trying to fix this problem (it boils down to properly preparing the b0 and passing it through S0). Now, we can tackle this in a more principled way.
- After some work for the past week, DKI seems to be in good shape; it's just the prediction runtime that we need to figure out.
- The work on GQI is also showing: it shows an unbeatable trade-off between signal fidelity and speed (esp. for DSI)
- The GPR is also tested! And it works (including the new multi-shell kernel). In #457, we fixed the gradients and modeled noise like Andersson, and then we finally completed the long-overdue #175.

In really excited about this PR. The price: this overly long message below (which basically says the right things explicitly, but I doubt anyone would have time to read it through carefully), and the sometimes verbose code (I'm trying to deal with that part right now, will lift the draft marker when done). But in summary, this is solid progress, folks.

---

## Summary

Adds a documentation **prediction gallery**: for several real dMRI datasets spanning the acquisition-scheme spectrum, it renders the *predicted* volumes produced by each applicable model in both **LOVO** (leave-one-volume-out, holding out the predicted orientation) and **single-fit** (fit once on all volumes) modes, and records a **combinatorial coverage manifest** of what was exercised and what was skipped (and why).

The whole gallery is driven by a single declarative `(dataset × model × mode)` matrix (`nifreeze._gallery`). A model's applicability is decided by a new **capability contract** on the model classes, so the issue's suggested coverage (DTI → {average, DTI}; HARDI → + GQI; multi-shell → {DKI, GQI, GPR}; DSI → GQI) falls out automatically, and the difference between "attempted" and "applicable" is exactly the skip manifest.

Resolves: #458

## Datasets (OpenNeuro, one subject per scheme, fetched via DataLad)

| Scheme | Accession | Acquisition |
| --- | --- | --- |
| Legacy DTI | ds000206 | 30 dir @ b=1000 (traveling phantom) |
| Single-shell HARDI | ds000114 | 64 dir @ b=1000 |
| Multi-shell | ds003138 | b=1000/2000/3000, merged from three separate shell files |
| DSI | ds004737 | compressed-sensing q-space (HASC92) |

Minimal preprocessing only: a median-otsu brain mask + b0 extraction, cropped to the brain bounding box. Two labels are stated plainly on the page: ds000114 is single-shell at b=1000 (high angular resolution, milder than textbook high-b HARDI), and ds004737 is compressed-sensing DSI, not a full 258-point grid.

## Visualization

Each cell renders a `nilearn` montage of six axial cuts (orientation-aware, fixed to the high-mass **cerebrum** and identical across every panel on a page) in five rows:

- **observed** and **predicted** (intensity-matched),
- **checkerboard** — observed/predicted interleaved so mismatches show as seams,
- **difference (relative)** — the in-mask residual normalized by the observed dynamic range (a fraction of signal), with a colorbar,
- **local correlation** — observed-vs-predicted Pearson correlation in a 7³ **sliding window** (stride 1, full resolution), on a bilateral [−1, 1] colormap with a colorbar.

Each panel title reports the **fit and predict times**; each page states the exact **OpenNeuro subject/run** used. GP cells additionally plot the fitted **angular covariance** function (Andersson 2015, Fig. 3) extracted from the kriging kernel. Figures render at 150 dpi.

## Also fixes a latent GP bug

`GPModel.fit_predict` was dead/broken: `BaseDWIModel._fit` builds `Model(gtab).fit(data)` (the DIPY convention), but `GaussianProcessModel` is constructed from a `kernel_model` and fits `(data, gtab)`. Every end-to-end GP test constructs `GaussianProcessModel` directly, so the wrapper path was never exercised. This PR overrides `_fit`/`fit_predict` in `GPModel` (vectorized over voxels, no chunking) so `GPModel(dataset).fit_predict(index)` works, and registers GP in `ModelFactory` (`"gp"`/`"gpr"`/`"gaussianprocess"`).

## What's included

- **Capability contract** (`supports_single_fit`, `applicable_schemes`, `requires_multishell`, `excludes_b0`) on the model classes, replacing scattered runtime guards.
- **Gallery harness** (`nifreeze._gallery`): matrix + capability filtering, nilearn renderer + covariance plot, JSON/Markdown coverage manifest, CLI (`python -m nifreeze._gallery.run`).
- **OpenNeuro loaders** with DataLad fetch, masking, cropping, and multi-shell merge.
- **Four thin nbsphinx notebooks** + `docs/gallery.rst`, wired with `nbsphinx_execute = "never"` so the fast, mocked PR docs build renders stored outputs (never executes model fits at build time). Legacy top-level notebooks are excluded from nbsphinx to keep `-W` clean.
- **Scheduled CI** (`.github/workflows/gallery.yml`, `tox -e gallery`, `gallery` extra): fetches the data, runs the matrix under coverage, executes the notebooks, uploads the manifest/panels/coverage, and — on manual dispatch — opens a PR refreshing the notebook outputs (scheduled runs only publish artifacts, to avoid GP run-to-run diff churn). No ANTs/AFNI needed — prediction only.

## Tests / validation

- `test/test_gallery.py`: synthetic-data matrix, capability gating, manifest round-trip, registry integrity, and a data-gated real-load test (skipped when data isn't cached).
- New model tests: the capability contract, GP factory resolution, and `GPModel.fit_predict`.
- Validated end-to-end on real data: ds000206 (all applicable models, incl. GP + covariance) and ds003138 (shell-merge → 114 volumes, multi-shell, DKI/GQI/DTI/GP-multishell all run).
- `make -C docs html` passes under `-W`.

## Notes for reviewers

Built on `main` (includes #175, so the multi-shell GP kernel is available). Notebook outputs are intentionally **not committed** here (they ship empty); the scheduled `gallery` job populates them. This is a **draft** pending that first CI refresh. X-Refs: #175.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/nifreeze/model/base.py
```diff
@@ -96,6 +105,21 @@ class BaseModel(ABC):
 
     __metaclass__ = ABCMeta
 
+    #: Declarative capability contract (read by callers such as the prediction
+    #: gallery to decide, *before* instantiation, whether a model applies to a
+    #: given dataset and mode). Subclasses override the class attributes below;
+    #: they are intrinsic model properties, centralizing constraints that were
+    #: otherwise only enforced by scattered runtime guards.
```
oesteban:
```suggestion
```

src/nifreeze/model/dmri.py
```diff
@@ -401,6 +401,10 @@ def fit_predict(self, index: int | None = None, **kwargs) -> Union[np.ndarray, N
 class AverageDWIModel(ExpectationModel):
     """A trivial model that returns an average DWI volume."""
 
+    # Shell-averaging needs the held-out index to pick the shell, so there is
+    # no single-fit (locked) mode (see :meth:`fit_predict`).
+    supports_single_fit = False
+
```
oesteban:
```suggestion
```

src/nifreeze/model/dmri.py
```diff
@@ -499,7 +511,63 @@ class GQIModel(BaseDWIModel):
...
+    convention that :class:`BaseDWIModel` implements: it is constructed from a
+    ``kernel_model`` and receives the gradient table at ``fit`` time, fitting all
+    voxels at once (vectorized across GP targets). ``_fit``/``fit_predict`` are
+    therefore overridden here rather than reusing the chunked DIPY path.
+    """
+
+    # The angular (+ optional radial) covariance targets orientation data; the
+    # applicable scheme further depends on ``kernel_model`` (single-shell for
+    # ``"spherical"``/``"exponential"``, multi-shell for ``"multishell"``), a
+    # distinction resolved by the caller (e.g. the gallery registry).
```
oesteban:
```suggestion
```

src/nifreeze/model/dmri.py
```diff
@@ -468,6 +468,10 @@ def fit_predict(self, index: int | None = None, *_, **kwargs) -> np.ndarray:
 class DTIModel(BaseDWIModel):
     """A wrapper of :obj:`dipy.reconst.dti.TensorModel`."""
 
+    # The tensor model is estimated from low-b data; DSI's dense q-space grid
+    # is out of scope.
```
oesteban:
```suggestion
```

src/nifreeze/model/dmri.py
```diff
@@ -482,13 +484,21 @@ class DTIModel(BaseDWIModel):
 class DKIModel(BaseDWIModel):
     """A wrapper of :obj:`~nifreeze.model.dki.DiffusionKurtosisModel`."""
 
+    # Kurtosis estimation needs >= 3 b-values (see ``check_multi_b`` guard in
+    # ``BaseDWIModel.__init__``); it only applies to multi-shell acquisitions.
```
oesteban:
```suggestion
```

src/nifreeze/model/base.py
```diff
@@ -96,6 +115,9 @@ class BaseModel(ABC):
 
     __metaclass__ = ABCMeta
 
+    supports_single_fit: bool = True
```
oesteban: Why keeping this? Shouldn't this be converted into canary?
```suggestion
    single_fit_is_canary: bool = False
    """Whether single-fit only makes sense as a self-consistency canary (see
    :meth:`_warn_single_fit_canary`)."""
```

src/nifreeze/model/dmri.py
```diff
@@ -263,6 +280,11 @@ def __init__(self, dataset: DWI, max_b: float | int | None = None, **kwargs):
 
         super().__init__(dataset, **kwargs)
 
+    def _warn_single_fit_canary(self) -> None:
+        """Warn if single-fit for this model is only a self-consistency canary."""
+        if self.single_fit_is_canary:
+            warnings.warn(SINGLE_FIT_CANARY_MSG, SingleFitCanaryWarning, stacklevel=3)
+
```
oesteban: Shouldn't this be in the base model foundation?

PR #468, opened 2026-07-23
ENH: Implement the locked fit feature for PET B-Spline model

jhlegarreta:
Implement the locked fit feature for the PET B-Spline model.

Add the corresponding tests.

### DIFF
--- a/src/nifreeze/model/pet.py
+++ b/src/nifreeze/model/pet.py
@@ -183,6 +183,8 @@ class BSplinePETModel(BasePETModel):
         "_n_ctrl": "Number of B-Spline control points",
     }
 
+    single_fit_is_canary = True
+
     def __init__(
         self,
         dataset: PET,
@@ -232,18 +234,28 @@ class BSplinePETModel(BasePETModel):
         the prediction for the start time.
         """
 
-        if index is None:
-            raise NotImplementedError(
-                "Fitting all frames at once (index=None) is not yet implemented. "
-                "Use LOVO mode by passing an integer index."
-            )
+        # Generate a time mask for the frames to fit
+        x_mask = np.ones(len(self._dataset), dtype=bool)
+        if self._locked_fit is None:
+            if index is None:
+                self._locked_fit = True
+                self._warn_single_fit_canary()
+            else:
+                # First call in LOVO mode: initialize as unlocked
+                self._locked_fit = False
+                x_mask[index] = False
+        elif index is None:
+            self._warn_single_fit_canary()
+            return None
+        elif self._locked_fit:
+            # Already fit on all frames; for held-out prediction keep full-fit mask
+            pass
+        else:
+            # Unlocked LOVO mode
+            x_mask[index] = False
 
         n_jobs = kwargs.pop("n_jobs", min(cpu_count() or 1, 8))
 
-        # Generate a time mask for the frames to fit
-        x_mask = np.ones(len(self._dataset), dtype=bool)
-
-        x_mask[index] = False if index is not None else x_mask[index]
         x = self._dataset.midframe[x_mask].tolist()
 
         # A.shape = (T, K - 4); t= n. timepoints, K= n. knots (with padding)
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -37,6 +37,7 @@ from nifreeze.data.dmri.utils import (
     DEFAULT_MAX_S0,
     DEFAULT_MIN_S0,
 )
+from nifreeze.data.pet import PET
 from nifreeze.model.base import (
     MASK_ABSENCE_WARN_MSG,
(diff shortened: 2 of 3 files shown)


jhlegarreta:
Fitting with all data is not implemented for PET so marked as a draft.

jhlegarreta:
@mnoergaard Is there a particular reason why the B-Spline fitting should not implement the single-fit (locked fit) feature?

oesteban:
> @mnoergaard Is there a particular reason why the B-Spline fitting should not implement the single-fit (locked fit) feature?

Sorry for chiming in. The only reason not to do it is that the fitting is so fast that it's not really worthwhile. But I agree we should implement it for completeness (and because it's not a very heavy lift).

jhlegarreta:
Re https://github.com/nipreps/nifreeze/pull/468#issuecomment-5157232159 It has been addressed @oesteban.

jhlegarreta:
I've realized that the `BasePETModel` model has this block that is not being called neither is present across other modalities:
https://github.com/nipreps/nifreeze/blob/ba8e32adf63405cd35cdad7be560c5cd06d5b9a2/src/nifreeze/model/pet.py#L167-L169

The `BaseModel` did also have it until removed in PR #62 commit 7322e93bc3eda2ebe5d2036b1f90c0f10596716c:
https://github.com/nipreps/nifreeze/pull/62/changes/7322e93bc3eda2ebe5d2036b1f90c0f10596716c#diff-59145ee0e5e3ec5769497e46f1bee4d47d3855cf4aa03a00e38d186394ff69f5L115

Since the PET models did not initially inherit from the base model, that has remained after the refactor.

There are some appearances of what the function checks:
https://github.com/search?q=repo%3Anipreps%2Fnifreeze+%22_locked_fit+is+not+None%22&type=code

I would defer to a separate PR removing the function or bringing it back to the base class, and using it in the above checks.

src/nifreeze/model/pet.py
```diff
@@ -183,6 +183,8 @@ class BSplinePETModel(BasePETModel):
         "_n_ctrl": "Number of B-Spline control points",
     }
 
+    single_fit_is_canary = True
+
```
oesteban:
```suggestion
```

I admit that the "canary" definition is really unclear so I apologize. The idea for "is canary" is that the model can fit the signal perfectly. Meaning, if you run "single fit", then you get exactly the data you fed in, with its noise and everything. That's why it is deemed a canary - if you run single fit and you don't get nearly 1.0 correlation between the target and the simulated images then something's broken.

That's not the case of this model, you'll not get ever the original frame, even if used in training.

src/nifreeze/model/pet.py
```diff
@@ -232,18 +234,28 @@ def fit_predict(self, index: int | None = None, **kwargs) -> Union[np.ndarray, N
         the prediction for the start time.
         """
 
-        if index is None:
-            raise NotImplementedError(
-                "Fitting all frames at once (index=None) is not yet implemented. "
-                "Use LOVO mode by passing an integer index."
-            )
-
-        n_jobs = kwargs.pop("n_jobs", min(cpu_count() or 1, 8))
```
oesteban, line 241: I would keep this here to have a cleaner diff

src/nifreeze/model/pet.py
oesteban: This is executed always, even if the fit was "locked".

src/nifreeze/model/pet.py
```diff
@@ -232,18 +234,28 @@ def fit_predict(self, index: int | None = None, **kwargs) -> Union[np.ndarray, N
...
+                x_mask[index] = False
+        elif index is None:
+            self._warn_single_fit_canary()
+            return None
+        elif self._locked_fit:
+            # Already fit on all frames; for held-out prediction keep full-fit mask
+            pass
+        else:
+            # Unlocked LOVO mode
+            x_mask[index] = False
```
oesteban, line 256: I believe this is not doing what it should be doing: the least squares fitting is going to be run with every prediction. The locked fit point is that you run it only once because you can (you're not holding any volume out)

src/nifreeze/model/pet.py
jhlegarreta: Not sure if I understand how the B-Spline model will be different from the e.g. GP on single/locked fit mode. If you fit a B-Spline on all available data points, you are constraining it to pass through those points, and thus, it will return those same points if you query them.

oesteban: > you are constraining it to pass through those points

That's the key, in the case of B-Splines you can't constrain it to pass through all the points. It's a LST fitting.

src/nifreeze/model/pet.py
jhlegarreta: Done in 9a47595.

PR #469, opened 2026-07-23
TST: Prefer using the DWI dataset generation fixture

jhlegarreta:
Prefer using the DWI dataset generation fixture: the `test_gpmodel_fit_predict` and `test_single_fit_canary_warning` test functions do not check whether the predictions match some expected value; rather, they verify that the LOVO and single-fit execution paths provide non-null and finite results, and emit warnings as necessary.

Avoids adding boilerplate code.


jhlegarreta:
@oesteban If there is a reason to require a particular distribution, then maybe we can modify the fixture so that we can specify the desired distribution?

oesteban (approved):

PR #471, opened 2026-07-23
STY: Prefer avoiding private import statements in tests

jhlegarreta:
Prefer avoiding private import statements in tests.

Improves consistency throughout the test by resolving names wherever they are needed, since only the `model` package is imported.


jhlegarreta:
Not sure if there was a reason for private import statements. We can also import all classes at the top if we prefer that over the need to fully resolve the names every time.

oesteban (approved):

PR #481, opened 2026-08-10
MNT: Add prediction gallery workflow badge to README

jhlegarreta:
Add prediction gallery workflow badge to README.


jhlegarreta:
This is a scheduled job, so its status can easily go unnoticed unless we incidentally realize while having a look at other builds. This may help in raising awareness.

PR #473, opened 2026-08-02
MNT: Require Python greater or equal to 3.12

jhlegarreta:
Require Python greater or equal to 3.12:
- Support for Python 3.10 will end in 2 months (EOL 2026-10).
- Commit b4f4f8f required DIPY `HEAD`, which requires Python >= 3.12.
- Update the minimum required versions for the dependencies accordingly.

Add Python 3.14 and 3.15 to the matrix of Python versions in CI testing workflow; add Python 3.14 to packaging CI workflow as Python 3.15 is still in pre-release status.

PR #485, opened 2026-08-13
MNT: Upper bound the `hatchling` version to fix `twine` incompatibility

jhlegarreta:
The new version of `hatchling` (1.32) generates metadata version 2.5, which is not yet supported by `twine`. Upper bound the `hatchling` version to fix the issue.

Fixes:
```
Run /tmp/baipp/bin/python -m twine check --strict /tmp/baipp/dist/*
/tmp/baipp/bin/python -m twine check --strict /tmp/baipp/dist/*
shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
env:
  UV_CACHE_DIR: /tmp/baipp-uv_cache_dir
  REQS_HASH:
    567a34dc4c0412a37952d1396403566f0d030748c60574977cdf16bcfdbca5da
  Checking
    /tmp/baipp/dist/nifreeze-26.0.0.dev221-py3-none-any.whl:
    ERROR    InvalidDistribution: Invalid
    distribution metadata: '2.5' is not a valid metadata version
```

raisd for example in:
https://github.com/nipreps/nifreeze/actions/runs/31725653046/job/94533249421

Documentation:
https://stackoverflow.com/a/79939707
https://github.com/pypa/hatch/issues/2292#issue-4559640411

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling", "hatch-vcs", "nipreps-versions"]
+requires = ["hatchling < 1.32", "hatch-vcs", "nipreps-versions"]
 build-backend = "hatchling.build"
 
 


kratsg:
Just a quick note @jhlegarreta  - newest twine supports it, but you might not have a recent enough python to do so. twine v7 seems ok for me.

jhlegarreta:
Thanks for weighing in @kratsg. I am not that proficient in the parts involved by the building process, so apologies if my arguments here are unrelated to your remark. We do no pin `twine`, so I assume the workflow is using the latest available version. As for Python, we require Python >= 3.10, so a version that is still maintained by the Python developers (PR #473 drops 3.10 in line with its EOL, and maybe we will drop 3.11 at the same time). That being said, I would be happy, if instead of upper bounding a dependency, lower bounding another one would fix the issue.

edgarrmondragon:
@jhlegarreta you need to bump hynek/build-and-inspect-python-package manually:

https://github.com/nipreps/nifreeze/blob/380dd43c8c9522d39373ae829eaddf0b52cd280d/.github/workflows/pypi-package.yml#L27

```diff
- - uses: hynek/build-and-inspect-python-package@v2
+ - uses: hynek/build-and-inspect-python-package@v3.0.1
```

Unfortunately Dependabot will not upgrade v2 -> v3.0.1 (see https://github.com/pypa/hatch/issues/2292#issuecomment-5282177778)

jhlegarreta:
I pushed and merged #486. Thanks both @kratsg and @edgarrmondragon.

